Issue #3, opened 2017-04-12
Support older versions of Node


vassyz:
Because the key is unwrapped using the `Buffer.from` method, older versions of Node (something like 5.1 or 5.3, around late 2015) will not be supported. I don't have time to look into this right now, but I can investigate once I get back from holiday.

rob-yoti:
@vassyz I only see the key unwrapped using `new Buffer` not `Buffer.from`.

Which file and line do you see 'Buffer.from` being used?

PR #5, opened 2017-04-18
Replace `Buffer.from` with `new Buffer`

rob-yoti:
`src/profile_service/index.js`

`Buffer.from` is not avaliable in Node < 5.10
https://nodejs.org/dist/latest-v7.x/docs/api/buffer.html#buffer_class_method_buffer_from_buffer

Changing it to `new Buffer` allows users with older node versions to use the SDK.

Fixes issue #3

### DIFF
--- a/src/profile_service/index.js
+++ b/src/profile_service/index.js
@@ -110,7 +110,7 @@ function decipherProfile(cipherText, key, iv, callback) {
 }
 
 function unwrapKey(wrappedKey, pem) {
-  let wrappedKeyBuffer = Buffer.from(wrappedKey, 'base64');
+  let wrappedKeyBuffer = new Buffer(wrappedKey, 'base64');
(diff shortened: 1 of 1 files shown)


vassyz:
`new Buffer` is deprecated, I'm not sure this is the right solution. This answer is pretty good: http://stackoverflow.com/a/14573049/1629709

rob-yoti:
I know `new Buffer` is deprecated, however it is what is used in every other part of this SDK. When @Alttaf added this small change with `new Buffer` it prevented anyone < 5.10.0 from using it.

vassyz:
I think the best solution for this issue would be to add a conditional similar to `typeof Buffer.from === "function"`. But for now, I think we should get this merged in, and change `new Buffer` to `Buffer.from` in a year (2018-04-01), when support for Node 4.x drops https://github.com/nodejs/LTS#lts-schedule1

PR #6, opened 2017-10-05
[SDK-115]: Added X-SDK identifier to HTTP header sent with API requests

echarrod:
Initial PR, once this is approved I can upload the new package, and then update the example project to use this new package. 
I also just added the instructions for installing the dependencies and running the tests.
@Alttaf - is bumping the version to 2.0.0 ok? Based on semantic versioning, I thought this extra header is an incompatible API change, and so 1.1.0 isn't appropriate.
I also just highlighted a part of the instructions in bold, as I missed this when I was scanning through, and it was quite important!

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,13 +79,13 @@ npm test
 ## Installing the SDK
 
 To import the Yoti SDK inside your project, you can use your favourite dependency management system.
-If you are using NPM, you can use `npm install -S -E yoti-node-sdk` to set the Yoti SDK as a dependency.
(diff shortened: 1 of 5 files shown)


vassyz:
It doesn't require a major version bump because it doesn't add breaking changes. I would go with version 2.0.0 based on it being published under a different name.

echarrod:
I haven't changed the name yet - we don't have a Jira ticket for it do we? I can add one, and check in that with this PR, so that they're merged into master at the same time

PR #7, opened 2017-10-11
[SDK-128]: Changed X-SDK to X-Yoti-SDK and updated package

echarrod:
I changed the README to have the version number as "yoti" : "x.x.x" to prevent having to change this each time. Would you say this makes sense? I'll update on npm, and then install to the example project once this is approved.


vassyz:
Looking through the Readme file, I think it needs a few tweaks. Let's talk.

PR #9, opened 2017-10-12
[SDK-128]: Updated example project to use new package

echarrod:
Also committed package-lock.json.

I'm thinking we should do this based on [npm docs](https://docs.npmjs.com/files/package-lock.json), [SO Answer 1](https://stackoverflow.com/questions/44297803/package-lock-json-role) and [SO Answer 2](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5)

package-lock.json file will be updated every time you use npm install since NPM 5.1. So we'd either be updating it every time before checking in, or discarding/ignoring it from source control every time.

Listed as the benefits are:

- Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
- Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
- To facilitate greater visibility of tree changes through readable source control diffs.
- And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

### DIFF
--- a/example/simple-login/package-lock.json
+++ b/example/simple-login/package-lock.json
@@ -1,0 +1,443 @@
+{
+  "name": "yoti-node-sdk-simple-login",
+  "version": "0.0.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "ansi-regex": {
+      "version": "2.1.1",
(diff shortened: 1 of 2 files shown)


vassyz:
Good find :+1: The SO Answer 2 link doesn't work.

echarrod:
Ah yes, I've updated it now with the correct link

PR #10, opened 2017-10-16
[SDK-131]: Updated License

echarrod:
Added format to JIRA ticket so that we can use this as a template for others


echarrod:
I'll add another commit to update the packag.json file with the new license

PR #14, opened 2017-11-20
[SDK-167]: Excluded test folder from published package using .npmignore

echarrod:
Nothing was required for example project due to the package.json file for that specifying private

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+example
+test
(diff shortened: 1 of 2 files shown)


vassyz:
Don't think this is working, installed using `npm install git://github.com:getyoti/yoti-node-sdk.git#npm-ignore` and it still adds the test older to my `node_modules`.

echarrod:
Good spot, I've corrected it now. Ran `npm pack` locally and did the same command you mentioned above to check it works as expected

PR #17, opened 2018-01-30
SDK-234 : Add Node signed request

sidibos:


src/yoti_request/index.js
```diff
@@ -0,0 +1,112 @@
...
+              request = superagent.put(`${server.configuration.connectApi}${endpoint}`);
+            } else {
+              request = superagent.patch(`${server.configuration.connectApi}${endpoint}`);
+            }
+
+            request
+                .query({nonce: nonce})
+                .query({timestamp: timestamp})
+                .query({appId: applicationId})
+                .query({payload: payloadString})
```
qzagarese: The payload should not go in the URL

src/yoti_request/index.js
```diff
@@ -0,0 +1,112 @@
...
+            break;
+
+          case 'DELETE':
+            // Build message to sign
+            messageToSign = `${httpMethod}&${endpoint}?nonce=${nonce}&timestamp=${timestamp}&appId=${applicationId}&payload=${payloadString}`;
+            request = superagent.put(`${server.configuration.connectApi}${endpoint}`)
+                .query({nonce: nonce})
+                .query({timestamp: timestamp})
+                .query({appId: applicationId})
+                .query({payload: payloadString})
```
qzagarese: same here

src/yoti_request/index.js
```diff
@@ -0,0 +1,112 @@
...
+    console.log('Signing the request message');
+    console.log('Payload ' + payloadString);
+
+    return new Promise((resolve, reject) => {
+        console.log('Making Http method ' + httpMethod + ' request');
+        // Initiate the right request
+        switch(httpMethod) {
+          case 'POST', 'PUT', 'PATCH':
+            // Build message to sign
+            messageToSign = `${httpMethod}&${endpoint}?nonce=${nonce}&timestamp=${timestamp}&appId=${applicationId}&payload=${payloadString}`;
```
qzagarese: please remove "&payload"; only concatenate "&" + the value of payloadString

src/yoti_request/index.js
```diff
@@ -0,0 +1,112 @@
...
+                .query({nonce: nonce})
+                .query({timestamp: timestamp})
+                .query({appId: applicationId})
+                .query({payload: payloadString})
+                .send(payloadJSON);
+            break;
+
+          case 'DELETE':
+            // Build message to sign
+            messageToSign = `${httpMethod}&${endpoint}?nonce=${nonce}&timestamp=${timestamp}&appId=${applicationId}&payload=${payloadString}`;
```
qzagarese: same as above. Also DELETE calls usually don't have a body

src/yoti_request/index.js
```diff
@@ -0,0 +1,112 @@
...
+                .query({timestamp: timestamp})
+                .query({appId: applicationId})
+                .query({payload: payloadString})
+                .send(payloadJSON);
+
+
+          default :
+            // Make default message request
+            // Build message to sign
+            messageToSign = `${httpMethod}&${endpoint}?nonce=${nonce}&timestamp=${timestamp}&appId=${applicationId}&payload=${payloadString}`;
```
qzagarese: same as above

src/yoti_request/index.js
```diff
@@ -0,0 +1,112 @@
...
+            request = superagent.get(`${server.configuration.connectApi}${endpoint}`)
+                .query({nonce: nonce})
+                .query({timestamp: timestamp})
+                .query({appId: applicationId})
+                .query({payload: payloadString});
```
qzagarese: same as above

sidibos:
@qzagarese I have applied those changes requested.

sidibos:
Also in POST, PUT, and PATCH request I am sending the payload as a JSON data `request.send(payloadJSON)` not sure if it's needed.

src/profile_service/index.js
```diff
@@ -1,11 +1,9 @@
 'use strict'
 
-const crypto = require('crypto');
+const yotiRequest = require('../yoti_request');
+const RequestPayload = require('../request_payload');
```
vassyz: Shouldn't we use `-` instead of `_` for name separator? To keep it consistent with `proto-root`

src/yoti_common/index.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const crypto = require('crypto');
+const forge = require('node-forge');
+
+exports.getRSASignatureForMessage = (message, pem) => {
+  let sign = crypto.createSign('RSA-SHA256');
```
vassyz: We should have a standard for indentation. It seems we're using 2 spaces 4 spaces and tabs throughout the project. Maybe we should address this in a future PR.

src/yoti_request/index.js
```diff
@@ -0,0 +1,102 @@
...
+    let timestamp =  Date.now();
+    let sdkIdentifier = 'Node';
+    let messageToSign;
+    let request;
+    let payloadString = Payload.getByteArray();
+    let payloadJSON = JSON.stringify(Payload.getRawData());
+    let endpointPath = `${server.configuration.connectApi}${endpoint}?nonce=${nonce}&timestamp=${timestamp}&appId=${applicationId}&${payloadString}`;
+
+
+    console.log('Signing the request message');
```
vassyz: I'd rather not have these `console.log`s here.

src/profile_service/index.js
```diff
@@ -37,93 +32,28 @@ ActivityDetails.prototype = {
...
+  return new Promise((resolve, reject) => {
+    yotiRequest.makeRequest(httpMethod, endpoint, pem, applicationId, Payload)
+        .then(response => {
+          if(response) {
+            let receipt = response.getReceipt();
+            let parsedResponse = response.getParsedResponse();
+            let decryptedProfile = yotiCommon.decryptCurrentUserReceipt(receipt, pem);
+            return resolve(new ActivityDetails(parsedResponse, decryptedProfile));
+          }
+          else {
```
vassyz: This `else` is not needed.

src/profile_service/index.js
```diff
@@ -37,93 +32,28 @@ ActivityDetails.prototype = {
...
+            console.log('error getting response data');
```
vassyz: Please make these two errors capitalised (`Error`).

src/yoti_common/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+exports.decryptCurrentUserReceipt = (receipt, pem, callback) => {
+  if(receipt.other_party_profile_content && Object.keys(receipt.other_party_profile_content).length > 0) {
+    let unwrappedKey = unwrapKey(receipt.wrapped_receipt_key, pem);
+    let decodedData = protoRoot.decodeEncryptedData(new Buffer(receipt.other_party_profile_content, 'base64'))
+    let iv = forge.util.decode64(decodedData.iv);
+    let cipherText = forge.util.decode64(decodedData.cipherText);
+
+    return decipherProfile(cipherText, forge.util.decode64(unwrappedKey), iv);
+  } else {
```
vassyz: Else not needed, improve error message.

src/yoti_common/index.js
```diff
@@ -0,0 +1,61 @@
...
+exports.getAuthKeyFromPem = (pem) => {
+  let privateKey = forge.pki.privateKeyFromPem(pem);
+  let publicKey = forge.pki.setRsaPublicKey(privateKey.n, privateKey.e);
+  let subjectPublicKeyInfo = forge.pki.publicKeyToAsn1(publicKey);
+  let p12Der = forge.asn1.toDer(subjectPublicKeyInfo).getBytes();
+  let p12b64 = forge.util.encode64(p12Der);
+  return p12b64;
+}
+
+exports.decryptCurrentUserReceipt = (receipt, pem, callback) => {
```
vassyz: Please move these methods at the top, and I'm not sure what the purpose of the `callback` argument is. I think it can be removed.

src/yoti_request/index.js
```diff
@@ -0,0 +1,101 @@
...
+              request = superagent.patch(endpointPath);
+            }
+
+            request.send(payloadJSON);
+            break;
+
+          case 'DELETE':
+            // Build message to sign
+            messageToSign = `${httpMethod}&${endpoint}?nonce=${nonce}&timestamp=${timestamp}&appId=${applicationId}`;
+            request = superagent.delete(`${server.configuration.connectApi}${endpoint}`)
```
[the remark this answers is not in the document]
sidibos: Because delete doesn't need payload, and the way I am sending the payload doesn't work when I use query - there is no key.

src/yoti_request/index.js
```diff
@@ -0,0 +1,101 @@
...
+                .query({nonce: nonce})
+                .query({timestamp: timestamp})
+                .query({appId: applicationId});
+
```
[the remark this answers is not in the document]
sidibos: Good spot that one.

src/yoti_request/index.js
```diff
@@ -0,0 +1,101 @@
+'use strict'
+
+const uuid = require('uuid');
+const superagent = require('superagent');
+const server = require('../../config').server;
```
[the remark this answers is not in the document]
sidibos: `.server` is needed here otherwise it won't work. We are specifically exporting server service from config module.

vassyz:
Included in #19

PR #19, opened 2018-02-14
SDK-248 : Add Node support for AML service

sidibos:


src/aml_service/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+  let PayloadObj = new Payload(amlProfile.getData());
+
+  return new Promise((resolve, reject) => {
+    yotiRequest.makeRequest(httpMethod, endpoint, pem, appId, PayloadObj)
+        .then(response => {
+          try {
+            // This will throw an error if the error message is included in the response.
+            AmlResultClass.checkAmlError(response.getParsedResponse());
+            // Check if all expected attributes are included in the result.
```
echarrod: I'm a fan of minimising comments - do we need this one here?

src/yoti_entity/aml.address.js
```diff
@@ -0,0 +1,73 @@
+'use strict'
+
+// AML address attributes
```
echarrod: Same with this - I would say this is apparent since it's in aml.address.js?

src/yoti_entity/aml.profile.js
```diff
@@ -0,0 +1,113 @@
+'use strict'
+
+// AML profile attributes
```
echarrod: ditto

src/yoti_request/aml.result.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
+
+// AML result attributes
```
echarrod: ditto

echarrod:
Also assuming you're going to squash and merge these commits into one for the base branch?

src/aml_service/index.js
sidibos: Yes I think I will keep that one - so it's clear the error raised is coming from the response.
The second one `Check if all expected attributes are included in the result.` could be removed if we think it's not needed.

README.md
```diff
@@ -171,6 +175,59 @@ The `profile` object provides a set of attributes corresponding to user attribut
...
+Yoti provides an AML (Anti Money Laundering) check service to allow a deeper KYC process to prevent fraud. This is a chargeable service, so please contact [sdksupport@yoti.com](mailto:sdksupport@yoti.com) for more information.
+
+Yoti will provide a boolean result on the following checks:
+* PEP list - Verify against Politically Exposed Persons list
+* Fraud list - Verify against  US Social Security Administration Fraud (SSN Fraud) list
+* Watch list - Verify against watch lists from the Office of Foreign Assets Control
+
+To use this functionality you must ensure:
+* Your application is assigned to your Organisation in the Yoti Dashboard - please see [here](https://www.yoti.com/developers/documentation) for further information.
+* Within your application please ensure that you have selected the 'Given Name(s)' and 'Family Name' attributes from the Data tab. This is the minimum requirement for the AML check.
```
echarrod: This part is slightly different to Seb's same text for [Ruby](https://github.com/getyoti/yoti-ruby-sdk/pull/14/files#diff-04c6e90faac2675aa89e2176d2eec7d8R201)

README.md
```diff
@@ -171,6 +175,59 @@ The `profile` object provides a set of attributes corresponding to user attribut
...
+
+The AML check uses a simplified view of the User Profile.  You need only provide the following:
+* Given Name(s)
+* Family Name
+* Country of residence - you will need to collect this from the user yourself
+
+To check a US citizen, you must provide two more attributes in addition to the three above:
```
echarrod: Ditto

sidibos:
@echarrod like always those commits will be squashed.

src/yoti_request/aml.result.js
sidibos: I also prefer to keep those comments there - might not be apparent for some people.

src/aml_service/index.js
echarrod: That's the one I was thinking, first one is fine

README.md
sidibos: Yes Seb changed the doc after my commit.

README.md
```diff
@@ -171,6 +175,60 @@ The `profile` object provides a set of attributes corresponding to user attribut
...
+**You must ensure you have user consent *before* using this service.**
+
+### Code Example
+
+Given a YotiClient initialised with your SDK ID and KeyPair (see [Configuration](#configuration)) performing an AML check is a straightforward case of providing basic profile data.
+
+```javascript
+
+// Initiate user profile data.
+let country = new YotiEntity.Country('GBR');
```
qzagarese: I'm a bit confused about the meaning of YotiEntity.
Also does Country need to be a type?

qzagarese: any thoughts @laurent-yoti

vassyz:
`Country` does not need to be a type. I also think `Obj` and `Entity` should not be used for naming. I'm not a huge fan of the word `perform` either.

src/aml_service/index.js
```diff
@@ -0,0 +1,86 @@
...
+  this.onWatchList = rawResult['on_watch_list'];
+}
+
+AmlResultObj.prototype = {
+  /**
+   * Check if user is a politically exposed person.
+   *
+   * @returns {exports.AmlResult.isOnPepList}
+   */
+  isOnPepList() {
```
laurent-yoti: Could use a getter if this AmlResultObj was a class, same for the field below. Would avoid the function call `amlResult.isOnPepList()` and become `amlResult.isOnPepList`

src/aml_service/index.js
```diff
@@ -0,0 +1,86 @@
+'use strict'
+
+const yotiRequest = require('../yoti_request');
+const AmlResultClass = require('../yoti_request/aml.result').AmlResult;
+const Payload = require('../yoti_request/payload').Payload;
+
+const AmlResultObj = function(rawResult){
+  this.onPepList = rawResult['on_pep_list'];
```
laurent-yoti: String 'on_pep_list' and others could be extracted as const since this pattern is used elsewhere, and also, the same string reused in the `getData()`

src/aml_service/index.js
```diff
@@ -0,0 +1,86 @@
...
+      'on_watch_list': this.isOnWatchList()
+    };
+  },
+
+  toString () {
+    return JSON.stringify(this.getData());
+  }
+}
+
+exports.performAmlCheck = (amlProfile, pem, appId) => {
```
vassyz: This might need a rewrite, but `performAmlCheck` should not require `pem` and `appId`.

src/aml_service/index.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  let PayloadObj = new Payload(amlProfile.getData());
+
+  return new Promise((resolve, reject) => {
+    yotiRequest.makeRequest(httpMethod, endpoint, pem, appId, PayloadObj)
+        .then(response => {
+          try {
+            // This will throw an error if the error message is included in the response.
+            AmlResultClass.checkAmlError(response.getParsedResponse());
```
laurent-yoti: Could probably factor out `response.getParsedResponse()`

src/yoti_request/index.js
```diff
@@ -0,0 +1,82 @@
+'use strict'
+
+const uuid = require('uuid');
+const superagent = require('superagent');
+const server = require('../../config').server;
+const yotiCommon = require('../yoti_common');
+
+let supportedMethods = ["POST", "PUT", "PATCH", "GET", "DELETE"];
```
vassyz: Use single quotes.

src/yoti_request/index.js
```diff
@@ -0,0 +1,82 @@
...
+exports.makeRequest = (httpMethod, endpoint, pem, appId, Payload) => {
+
+    // Make sure payload is an object
+    if(!Payload) {
+      throw new Error('Payload should be an object of type Request/Payload');
+    }
+
+    // Check if request method is supported
+    if(supportedMethods.indexOf(httpMethod) === -1) {
+      throw new Error('Http method ' + httpMethod + ' is not supported');
```
vassyz: HTTP

src/yoti_request/index.js
```diff
@@ -0,0 +1,82 @@
...
+                  let parsedResponse = JSON.parse(response.text);
+                  let receipt = parsedResponse.receipt;
+                  return resolve(new YotiResponse(parsedResponse, receipt));
+              }
+              catch (err) {
+                return reject(err);
+              }
+          })
+          .catch((err) => {
+              console.log('Error getting data from connect api: ' +  err.message);
```
vassyz: Connect API

README.md
lolodede: Feels like the term YotiEntity is here to act as an umbrella package domain name.
The code could just read something like:
```
import {Country} from YotiEntity
const country = new Country(...)
```
Yet not sure that Country is really related to Yoti, and then what is a YotiEntity?

src/aml_service/index.js
echarrod: We need appId don't we? 
I was working from `/aml-check?appId={app_id}&timestamp={timestamp}&nonce={nonce}` from https://lampkicking.atlassian.net/wiki/spaces/IPASS/pages/139068116/Connect+API+AML+Endpoints

vassyz: Yes, but the request should already know what the value of appId is. There's no point sending it from all the services that make a request. The use of appId is debatable, but pem definitely does not belong there.

src/yoti_request/aml.result.js
```diff
@@ -0,0 +1,42 @@
...
+   * @param rawResult
+   */
+  static checkAttributes (rawResult) {
+    if (!rawResult instanceof Array) {
+      throw new Error('Result Data should be an array');
+    }
+
+    let expectedElements = [ON_PEP_LIST_ATTR, ON_FRAUD_LIST_ATTR, ON_WATCH_LIST_ATTR];
+
+    for (let key in expectedElements) {
```
vassyz: Never use `for in` to loop through Arrays in JavaScript. You can replace it with something like `for (let i = 0; i < expectedElements.length; i++) {`

sidibos: Never? the difference I know is `for ...in` syntax only returns index of elements that have value in it. If there are other reasons not to use it, it will be good to share them.

src/aml_service/index.js
sidibos: This is following the same approach used in `ActivityDetails.getReceipt`, I don't see anything wrong with it. And I don't see why loading `pem` file in some kind of global configuration will make it any better.

src/aml_service/index.js
```diff
@@ -0,0 +1,86 @@
+'use strict'
+
+const yotiRequest = require('../yoti_request');
+const AmlResultClass = require('../yoti_request/aml.result').AmlResult;
```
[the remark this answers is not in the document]
sidibos: I believe '.' is used in `javascript` file name instead of snake_case or camelCase, I followed the same approach used in `src/proto-root`.
Also I am using `yoti_` for name-spacing as these are part of `Yoti` package, they might not belong to `Yoti`.

src/aml_service/index.js
sidibos: I am not sure why this should be refactored out.

src/aml_service/index.js
sidibos: In fact these are already declared as a `const` in `yoti_request/aml.result`, I will probably move them inside the `common` file and use them across the package.

src/aml_service/index.js
sidibos: I used a function here to be consistent with the other SDKs (PHP, Java) as this is the interface that clients will be using. But I am happy to use an attribute instead.

src/aml_service/index.js
sidibos: Ok I see what you mean, put it in a var instead.

README.md
sidibos: I am using `yoti` as a namespace as these are part of `yoti` package, and `entity` is a way to group those elements together, so I am following the folder name `yoti_entity`. I can understand `entity` is not widely used in `node` community, I can replace that folder name with `aml_type` or `aml_item`? Let me know your thought please.

src/aml_service/index.js
sidibos: Also `Country` doesn't need to be a type, I used a type here in case we need to add more `Country` attribute(s) in the future like `country_name`. But I am happy to squeeze this into `AmlAddress` as an attribute `countryCode`.

src/yoti_request/aml.result.js
vassyz: Yes, never, because it's not meant to do that. It iterates over the enumerable properties of an object. https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea

src/aml_service/index.js
sidibos: ES6 is already in the current version of this package which is already live. So i think that should be fine. 
Also regarding the signed-request, I move it and other functions inside `common` with the idea of sharing them across this package, so if in the future there is any common module that includes it, it should be easy to replace it with that.

src/yoti_request/aml.result.js
sidibos: I could change that, although it wouldn't have affected the current implementation since I am using `if (!rawResult.hasOwnProperty(attr)) {`.

vassyz: The code works, as long as somewhere in the code or a library a property is not being added to an object. Then it breaks.

sidibos:
I have applied some changes taking in count the reviews, please have a look.

src/aml_service/index.js
```diff
@@ -0,0 +1,64 @@
...
+const httpRequest = require('../request');
+const AmlResultClass = require('../request/aml.result').AmlResult;
+const {Payload} = require('../request/payload');
+const constants = require('../yoti_common/constants');
+const ON_PEP_LIST_ATTR = constants.ON_PEP_LIST_ATTR;
+const ON_FRAUD_LIST_ATTR = constants.ON_FRAUD_LIST_ATTR;
+const ON_WATCH_LIST_ATTR = constants.ON_WATCH_LIST_ATTR;
+
+const AmlResult = function(rawResult){
+  this.isOnPepList = rawResult[ON_PEP_LIST_ATTR];
```
vassyz: The way we are using constants here feels more confusing than helpful to me. Can I get some more opinions @echarrod @qzagarese @laurent-yoti 
I would just use `rawResult['on_pep_list'];`

src/aml_service/index.js
vassyz: The request itself can receive pem and appId arguments. These values are global, they might be used outside of this request, that's why it's better using them in a configuration object. You can see that an AML check does not require a pem file, but the request will ask for one. This is not the correct behaviour.

src/aml_type/aml.address.js
```diff
@@ -0,0 +1,69 @@
...
+    let data = {};
+    data[POSTCODE_ATTR] = this.postcode;
+    data[COUNTRY_ATTR] = this.countryCode;
+
+    return data;
+  }
+
+  checkCountryCode () {
+    if(this.countryCode === '') {
+      throw new Error('CountryCode cannot be empty');
```
vassyz: Is it also worth checking that the country code doesn't have 3 letters?

src/aml_type/aml.profile.js
```diff
@@ -0,0 +1,112 @@
...
+  getSsn () {
+    return this.ssn;
+  }
+
+  /**
+   * Check if amlAddress is an object.
+   *
+   * @param amlAddress
+   */
+  static checkAmlAddress (amlAddress) {
```
vassyz: Does using `validate` instead of `check` make more sense?

src/aml_type/aml.address.js
sidibos: imo it's better to keep this check in one place, `Connect API` is already checking this an returning the error. It sounds like duplicate to me. What is Connect API changes the error message or starts accepting 2 letters, we will have to change it here as well.

sidibos: AML check do requires `pem` file as it is calling `signedRequest` module to sign the message and using `request` module to send the request.
Please give your opinion @qzagarese, @laurent-yoti

src/aml_type/aml.profile.js
sidibos: either way is fine by me.

src/aml_type/aml.address.js
sidibos: Also `request` and `getRSASignatureForMessage` module shouldn't be tight to some configuration variables. If these were to move into a `Yoti` package that will be shared across the company. It will be impossible to achieve this scenario. And right now it's shared by `ActivityDetails` and `AML check`. 
These are just reusable dependencies/modules that shouldn't rely on any other element to function.

src/aml_service/index.js
echarrod: Can you just do `rawResult[constants.ON_FRAUD_LIST_ATTR];`, which removes the need for declaring the extra intermediary variables above, but also doesn't declare the string constant in the method

src/aml_type/aml.profile.js
echarrod: I prefer `validate`, `check` might just be checking for its presence

src/aml_service/index.js
sidibos: Yes that could do, it just seemed a bit long to me hence I used the declaration above.

sidibos: As long as it's not confusing to any one else.

src/aml_type/aml.address.js
vassyz: Please add basic testing and maybe we can get another general review from @laurent-yoti 
Also, make sure you update from master.

vassyz:
Please add basic testing and maybe we can get another general review from @laurent-yoti
Also, make sure you update from master.

src/request/payload.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Get payload as a JSON string.
+   *
+   * @returns {string}
+   */
+  getPayloadJSON() {
+    let data = this.data;
+    if(typeof data === 'string') {
```
vassyz: What is the purpose of this check?

sidibos:
LGTM - if it stays this way.

PR #20, opened 2018-03-29
SDK-229: Update image helper

vassyz:
Not entirely happy with this as it still requires a base64 conversion to save the image. I was unable to save a buffer object to an image file in Node. Let me know your thoughts. I also made some improvements to the way SDK configuration works.


src/proto-root/utils.js
```diff
@@ -2,32 +2,36 @@
...
       return `data:image/jpeg;base64,${value.toBase64()}`;
-    // Date as string in RFC3339 format (YYYY-MM-DD).
-    case DATE:
-      return value.toUTF8();
     // PNG indicates a standard .png image.
     case PNG:
       return `data:image/png;base64,${value.toBase64()}`;
     default:
-      return value;
+      return `data:image/jpeg;base64,${value.toBase64()}`;
```
echarrod: Why do we treat this as a JPEG?

vassyz: No reason, just assuming the default image type is JPEG. the default branch shouldn't exist, but I have to return something because of the arrow function. What would you prefer? `return value`? `return value.toBase64()`.

echarrod: Yes I guess I was thinking it should never get to this point if it is a JPEG, I think `value` would be best, but `value.toBase64` also ok

src/profile_service/index.js
```diff
@@ -21,6 +21,7 @@ ActivityDetails.prototype = {
   getUserId() { return this.receipt.remember_me_id; },
   getUserProfile() { return this.profile; },
   getOutcome() { return this.receipt.sharing_outcome; },
+  getBase64SelfieUri() { return this.profile.base64SelfieUri; },
 };
 
 const payload = new Payload('');
```
sidibos: Unrelated to this commit - is this right `console.log(`Error getting response data: $${err.message}`);` - 2 `$$`?

vassyz: Yup, I'll change it 👍

sidibos:
👍

PR #21, opened 2018-04-06
SDK-280 : Add age verified helper

sidibos:
I will add a test case for this after this review if needed.

### DIFF
--- a/src/profile_service/index.js
+++ b/src/profile_service/index.js
@@ -2,8 +2,8 @@
 
 const httpRequest = require('../request');
 const yotiCommon = require('../yoti_common');
-const Payload = require('../request/payload').Payload;
 const Age = require('../yoti_common/age').Age;
+const Payload = require('../request/payload').Payload;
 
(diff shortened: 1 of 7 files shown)


src/profile_service/index.js
```diff
@@ -15,6 +16,8 @@ const ActivityDetails = function main(parsedResponse, decryptedProfile) {
     acc[propName] = current[propName];
     return acc;
   }, {});
+  const age = new Age(this.profile);
+  this.profile.isAgeVerified = age.isVerified();
```
vassyz: Can't we add a conditional here `if (age.isVerified() !== null)`? It should not add the property if age verification does not exist. Adding this null property breaks the tests as well.

sidibos: I have double checked I can see it's not adding the attribute to the profile if the value is not selected in the Dashboard - I could add an if there.

vassyz: What do you mean by consistent? The other attributes are not being sent if they are null.

vassyz:
Looks good, can we also update `browser.js` within this PR?

PR #24, opened 2018-05-24
SDK-364-profile-anchors: Add anchor parsing

sidibos:
I will add the documentation and tests to this PR later.


README.md
```diff
@@ -287,7 +287,10 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
     * [X] Address `postalAddress`
     * [X] Gender `gender`
     * [X] Nationality `nationality`
-  * [X] Base64 Selfie URI `.getBase64SelfieUri()`
+  * [X] UserProfile `.getProfile()`
+    * [X] Full Name `.getFullName().getValue()`
+    * [X] Full Name - Source `.getFullName().getSources()[0].getValue()`
```
echarrod: Since there can be multiple sources, do you think we should either mention that, or show a way of retrieving all of the sources here? (obviously same below with verifiers)

vassyz: I think we should remove this list altogether and have it properly displayed in the documentation.

src/profile_service/profile.js
```diff
@@ -0,0 +1,277 @@
...
+module.exports.Profile = class Profile {
+  constructor(profileData) {
+    this.givenNames = null;
+    this.familyName = null;
+    this.fullName = null;
+    this.postalAddress = null;
+    this.gender = null;
+    this.emailAddress = null;
+    this.phoneNumber = null;
+    this.phoneNumber = null;
```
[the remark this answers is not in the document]
sidibos: Yes it looks like - but the code was throwing an error when those properties were called in the setter and needed to be initialised in the constructor.
I am not getting the same error now when I remove them from the constructor - So I can remove them.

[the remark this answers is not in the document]
vassyz: I was talking about this line: `this.phoneNumber = null;`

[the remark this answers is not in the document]
sidibos: 👍

vassyz:
@sidibos could you make sure `npm run lint` passes without errors? It's definitely easier if you add an eslint plugin to your editor.

sidibos:
I am planning to run `lint` after I finish adding the test cases.

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,152 @@
...
+                const serverX509Certs = AnchorProcessor.convertCertsListToX509(originServerCerts);
+                originAnchorObj.originServerCerts = serverX509Certs;
+                anchorsData[key].push(new AttributeAnchor(originAnchorObj));
+                anchorsData.processedValues[key].push(originAnchorObj.value);
+              }
+            }
+          }
+        }, anchorsData); // End Object loop
+      } // End for loop
+    } // End for loop
```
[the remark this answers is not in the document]
sidibos: When you are looking at this in a small window `bracket highlight` wouldn't help.
But I am happy to remove them.

src/profile_service/index.js
```diff
@@ -25,6 +32,7 @@ const ActivityDetails = function main(parsedResponse, decryptedProfile) {
 ActivityDetails.prototype = {
   getUserId() { return this.receipt.remember_me_id; },
   getUserProfile() { return this.profile; },
+  getProfile() { return this.userProfile; }, // This is the new profile object
```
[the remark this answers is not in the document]
sidibos: Yes a bit confusing, but I am using `getProfile()` as this is the name we will be using in the future and `getUserProfile()` will be removed.
I can change the variable name from `this.userProfile` to `this.extendedProfile`. 
Although I think it's better to keep the function as it is, which will help in the transition from `getUserProfile()` to `getProfile()`.

src/yoti_common/anchor.processor.js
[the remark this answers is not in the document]
echarrod: I'm just thinking whether we set a precedent - we wouldn't want to do this for every loop, but then where do we draw a line - when there's 2 or 3 brackets in consecutive lines? In my opinion it would be better to not have them at all. Feel free to leave as is though if you think it helps a lot

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,176 @@
...
+          const oidIndex = AnchorProcessor.findOidIndex(extensionsData, { id: anchorTypes[key] });
+          if (oidIndex !== -1) {
+            const anchorExtension = extensionsData[oidIndex];
+            const anchorValue = anchorExtension.value;
+            // Convert Anchor value from ASN.1 format to an object
+            const extensionObj = forge.asn1.fromDer(anchorValue.toString('binary'));
+            if (extensionObj) {
+              const anchorParsedValue = extensionObj.value[0].value;
+              // Make sure the anchor values are unique
+              if (!anchorsData.processedValues[key].includes(anchorParsedValue)) {
```
qzagarese: You definitely do not want to eliminate a PASSPORT anchor if there is an existing one as they might have different subType values. I guess you need to remove this.
If the goal of processedData was only to include the list of anchors without duplicates, then you can get rid of it.

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,176 @@
...
+
+    let anchorObj = {
+      value: '',
+      subType: '',
+      signedTimeStamp: '',
+      originServerCerts: '',
+    };
+
+    for (let i = 0; i < anchors.length; i += 1) {
+      const protoAnchor = anchors[i];
```
qzagarese: This section has gotten a bit difficult to read. Can we extract a function that processes one anchor at the time and returns what you would push into `anchorsData.sources` or `anchorsData.verifiers`?

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,176 @@
...
+            const extensionObj = forge.asn1.fromDer(anchorValue.toString('binary'));
+            if (extensionObj) {
+              const anchorParsedValue = extensionObj.value[0].value;
+              // Make sure the anchor values are unique
+              if (!anchorsData.processedValues[key].includes(anchorParsedValue)) {
+                anchorObj.value = anchorParsedValue;
+                const originServerCerts = anchorObj.originServerCerts;
+                const serverX509Certs = AnchorProcessor.convertCertsListToX509(originServerCerts);
+                anchorObj.originServerCerts = serverX509Certs;
+                anchorsData[key].push(new YotiAnchor(anchorObj));
```
qzagarese: The function I suggested at line 57 should actually return the `YotiAnchor` you instantiate here.

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,176 @@
...
+
+    for (let i = 0; i < anchors.length; i += 1) {
+      const protoAnchor = anchors[i];
+      const certificatesList = protoAnchor.originServerCerts;
+
+      anchorObj = Object.assign(anchorObj, protoAnchor);
+      anchorObj.signedTimeStamp = this.processSignedTimeStamp(protoAnchor.getSignedTimeStamp());
+
+      for (let n = 0; n < certificatesList.length; n += 1) {
+        const certArrayBuffer = certificatesList[n];
```
qzagarese: The content of this for can also extracted into a function that either returns the value of the anchor or null.
The caller should loop over the certificates and set the value in the anchorObj only if it is not null.

tests/profile-service.profile.spec.js
```diff
@@ -0,0 +1,92 @@
...
+      const sourceValue = profileObj.getFullName().getSources()[0].getValue();
+      expect(sourceValue).to.equal(expectedValue);
+    });
+  });
+
+  describe('#getFullName#getVerifiers#getValue', () => {
+    it('should return full_name first verifier value', () => {
+      const expectedValue = 'YOTI_ADMIN';
+      const verifierValue = profileObj.getFullName().getVerifiers()[0].getValue();
+      expect(verifierValue).to.equal(expectedValue);
```
qzagarese: can you please add some assertions on the other fields of the anchor?

tests/yoti-common.anchor.processor.spec.js
```diff
@@ -0,0 +1,79 @@
...
+        const expectedSourceValue = 'PASSPORT';
+        const ppAnchorObj = parseAnchorData(ppSourceAnchor);
+        const anchorSources = AnchorProcessor.process([ppAnchorObj]);
+        const anchorValue = anchorSources.sources[0].getValue();
+
+        expect(anchorValue).to.equal(expectedSourceValue);
+      });
+    });
+
+    context('when processing duplicate Source Anchors', () => {
```
qzagarese: this is not expected behaviour, so you might want to get rid of this test and update the logic to allow duplicate values, as you can have duplicate values with different sub types (e.g. PASSPORT + NFC and PASSPORT + OCR)

tests/yoti-common.anchor.processor.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+    context('when processing Verifier Anchor data', () => {
+      const verifierAnchor = fs.readFileSync('./tests/sample-data/yoti-common/verifier-anchor.txt', 'utf8');
+
+      it('should return YOTI_ADMIN as value', () => {
+        const expectedAnchorValue = 'YOTI_ADMIN';
+        const AnchorObj = parseAnchorData(verifierAnchor);
+        const anchorVerifiers = AnchorProcessor.process([AnchorObj]);
+        const anchorValue = anchorVerifiers.verifiers[0].getValue();
+
```
qzagarese: As you are loading a real anchor, you might want to add assertions about the other anchor fields.

src/yoti_common/anchor.processor.js
sidibos: Yes I forgot to remove that functionality, this was decided away back (not my idea though) to make the anchors values unique and then the decision was overturned later on.

tests/profile-service.profile.spec.js
sidibos: I have added assertions for other fields of the anchor in `yoti-common.anchor.processor.sepc.js`

src/yoti_common/anchor.processor.js
qzagarese: I know it wasn't your idea :-)

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,301 @@
...
+   * Get Oids anchors from a single certificate.
+   *
+   * @param certArrayBuffer
+   * @param subType
+   * @param yotiSignedTimeStamp
+   * @param X509Certs
+   *
+   * @returns {Array}
+   */
+  static getOidsAnchorsByCertificate(certArrayBuffer, subType, yotiSignedTimeStamp, X509Certs) {
```
qzagarese: I think this can be just `getAnchorsByCertificate` as the name should say what the function does, instead of how.
Much more readable now btw.

src/yoti_common/anchor.processor.js
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  /**
+   * Merge arrays by anchor type.
+   *
+   * @param targetList
+   * @param sourceList
+   * @returns {*}
+   */
+  static mergeAnchorsList(targetList, sourceList) {
```
qzagarese: `mergeAnchorsLists`

PR #28, opened 2018-09-14
SDK-541: Add the ability to retrieve DocumentDetails as Attribute value

sidibos:


src/data_type/document.details.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+
+const VALIDATION_PATTERN = /^([A-Za-z_]*) ([A-Za-z]{3}) ([A-Za-z0-9]{1}).*$/;
+const TYPE_INDEX = 0;
+const COUNTRY_INDEX = 1;
+const NUMBER_INDEX = 2;
+const EXPIRATION_INDEX = 3;
```
bucky-boy: Is there a concept of _private_ in PHP?
Should these be _private_?

src/yoti_common/attribute.converter.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const constants = require('./constants');
+const { DocumentDetails } = require('../data_type/document.details');
+
+module.exports.AttributeConverter = class AttributeConverter {
+  static convertValueBaseOnAttributeName(value, attrName) {
```
bucky-boy: convertValueBasedOnAttributeName

src/yoti_common/anchor.processor.js
```diff
@@ -55,7 +55,7 @@ module.exports.AnchorProcessor = class AnchorProcessor {
   static processSingleAnchor(anchorObj) {
     let anchorsList = this.getResultFormat();
 
-    if (!anchorObj instanceof Object) {
+    if (!(anchorObj instanceof Object)) {
```
bucky-boy: Was this a bug?

src/yoti_common/anchor.processor.js
```diff
@@ -205,7 +205,7 @@ module.exports.AnchorProcessor = class AnchorProcessor {
 
     for (let i = 0; i < anchorTypes.length; i += 1) {
       const anchorType = anchorTypes[i];
-      sourceList[anchorType].forEach(yotiAnchorObj => {
+      sourceList[anchorType].forEach((yotiAnchorObj) => {
```
bucky-boy: I though this commit was about DocumentDetails.... these changes appear to be for anchors.  What's going on here?

tests/data-type.attribute.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    });
+  });
+  context('Attribute.getName()', () => {
+    it('it should return document_details', () => {
+      expect(attributeObj.getName()).to.equal('document_details');
+    });
+  });
+  context('Attribute.getSources()[0]', () => {
+    it('it should return an Anchor object', () => {
+      expect(attributeObj.getSources()[0]).to.be.an('object');
```
bucky-boy: Can we do more than prove that its an object?

tests/data-type.attribute.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    });
+  });
+  context('Attribute.getVerifiers()[0]', () => {
+    it('it should return an Anchor object', () => {
+      expect(attributeObj.getSources()[0]).to.be.an('object');
```
bucky-boy: Test says _getVerifiers()_, but this is calling _getSources()_

tests/data-type.document.details.spec.js
```diff
@@ -0,0 +1,89 @@
...
+      }
+    });
+  });
+  context('when value is four words', () => {
+    it('it should parse one optional attribute', () => {
+      const documentDetails = new DocumentDetails('PASSPORT GBR 01234567 2020-01-01');
+      expect(documentDetails.type).to.equal('PASSPORT');
+      expect(documentDetails.issuingCountry).to.equal('GBR');
+      expect(documentDetails.documentNumber).to.equal('01234567');
+      expect(documentDetails.expirationDate.toISOString().slice(0,10)).to.equal('2020-01-01');
```
bucky-boy: What else is on the String if we do not _slice()_ it?

src/data_type/document.details.js
```diff
@@ -0,0 +1,61 @@
...
+    this.validateData(value);
+    const parsedValues = value.split(' ');
+
+    this.type = parsedValues[TYPE_INDEX];
+    this.issuingCountry = parsedValues[COUNTRY_INDEX];
+    this.documentNumber = parsedValues[NUMBER_INDEX];
+
+    if (parsedValues.length > EXPIRATION_INDEX) {
+      const dateValue = parsedValues[EXPIRATION_INDEX];
+      const dateObj = new Date(dateValue);
```
bucky-boy: What happens if _dateValue_ is not a correctly formatted date value?  Won't this fail?

src/data_type/document.details.js
sidibos: You mean in Node.js?
Not really, unless you use typescript.

src/yoti_common/attribute.converter.js
sidibos: Yes will fix that typo.

src/yoti_common/anchor.processor.js
sidibos: Not a bug, but this was recommended by ESLint.

src/yoti_common/anchor.processor.js
sidibos: Yes this is done from ESLint recommendation.
You can see in the commit message.

tests/data-type.attribute.spec.js
sidibos: Not in this case as I haven't defined a proper class for anchors, it's just a prototype.
So object is fine.

tests/data-type.attribute.spec.js
sidibos: Ok will fix that - thx.

tests/data-type.document.details.spec.js
sidibos: This one of the way to get the format the date in the short format `Y-d-m` in javascript.
If you don't slice it, you will get the full date as `2020-01-01T00:00:00.000Z`

src/data_type/document.details.js
sidibos: Nope it will return null, hence there is isNaN check on line 35.

src/yoti_common/anchor.processor.js
bucky-boy: Cool.  It is clearer this way

tests/data-type.attribute.spec.js
bucky-boy: Can we prove what's on the object?  The _type_, _subType_, _value_ etc...?

tests/data-type.document.details.spec.js
bucky-boy: Is there any other js _Date_ type that only applies to Dates?  I mean, without a time component?

src/data_type/document.details.js
bucky-boy: Fair enough

sidibos:
_We want to be sure that failing to parse a DocumentDetails attribute will not stop the profile and other attributes getting parsed.
Do we have a test somewhere that shows that?_
We don't have a test case for this, but the processing of Document details is inside a try catch, should any errors occur there, it wouldn't affect other attributes. 
If we want to test this use case, we will need a protobuf profile data whith an invalid DocumentDetails.

tests/data-type.attribute.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    });
+  });
+  context('Attribute.getSources()[0]', () => {
+    it('it should return an Anchor object', () => {
+      expect(attributeObj.getSources()[0]).to.be.an('object');
+      expect(attributeObj.getSources()[0].getValue()).to.equal('DRIVING_LICENCE');
+      expect(attributeObj.getSources()[0].getSubType()).to.equal('');
+      expect(attributeObj.getSources()[0].getSignedTimeStamp().getTimestamp()
+          .toUTCString()).to.equal('Wed, 11 Apr 2018 12:13:03 GMT');
+      expect(attributeObj.getSources()[0].getOriginServerCerts().signatureOid).to.equal('');
```
echarrod: Are there any other properties on the OriginServerCert that we can test against, that wouldn't be an empty string?

sidibos: I got around this, it shouldn't be empty.

PR #30, opened 2018-11-08
SDK-725 / Issue 31: Added support for dynamic scenarios

Alttaf:
This pr is to add dynamic attribute lists to the sdk, changes have been rebased on latest development branch and an extra on this pr is the addition of git pre commits which run the tests


README.md
```diff
@@ -22,18 +22,21 @@ How to initialise your configuration
 6) [Profile Retrieval](#profile-retrieval) -
 How to retrieve a Yoti profile using the token
 
-7) [Handling Users](#handling-users) -
+7) [Dynamic Policy Generation](#dynamic-policy-generation) -
+How to retrieve a qrcode link for dynamic policies
```
echarrod:
```suggestion
How to retrieve a QR Code link for dynamic policies
```

README.md
```diff
@@ -157,6 +160,34 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
 
 [Example reference](examples/profile/index.js)
 
+
+## Dynamic Policy Generation
+
+Dynamic policy generation is a way of requesting non-static policies for your application
```
echarrod: Full stops at the end of this sentence and the one below please 👍

README.md
```diff
@@ -157,6 +160,34 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
...
+This is useful when you want or need different permutations of attributes from Yoti
+
+E.g:
+ 
+* Request 1: full-name and date-of-birth
+* Request 2: full-name and address
+
+This service will query Yoti Core (via Connect) for a QR Code/link associated with the requested policy.
+
+This QR Code/link should then be embedded into your page using the Connect widget to begin a share with Yoti.
```
echarrod: Same WRT Connect

README.md
```diff
@@ -157,6 +160,34 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
...
+Dynamic policy generation is a way of requesting non-static policies for your application
+
+This is useful when you want or need different permutations of attributes from Yoti
+
+E.g:
+ 
+* Request 1: full-name and date-of-birth
+* Request 2: full-name and address
+
+This service will query Yoti Core (via Connect) for a QR Code/link associated with the requested policy.
```
echarrod: It would be good if @KiranBali could review this text - AFAIK `Yoti Core` (and probably `Connect` too) are not terms which clients will be aware of. It might be better if we just state `will query Yoti for a QR Code`.

src/client/index.js
```diff
@@ -34,4 +35,19 @@ module.exports.YotiClient = class YotiClient {
   performAmlCheck(amlProfile) {
     return amlService.performAmlCheck(amlProfile, this.pem, this.applicationId);
   }
+
+  /**
+   * @desc given a dynamic sharing request get a custom qr code denoted by the
```
echarrod:
```suggestion
   * @desc given a dynamic sharing request, get a custom QR Code denoted by the
```

src/client/index.js
```diff
@@ -34,4 +35,19 @@ module.exports.YotiClient = class YotiClient {
...
+   * policy stated in the request.
+   * @param {*} dynamicSharingRequest is the request that contains the policy.
+   * @returns {Promise} constaining a dynamic
```
echarrod: - Typo: `constaining` 
- Containing a dynamic...QR Code?

src/client/index.js
```diff
@@ -34,4 +35,19 @@ module.exports.YotiClient = class YotiClient {
...
+   */
+  getDynamicPolicyQRCodeLink(dynamicPolicyRequest) {
+    // Build a dynbanic service, that gets the service
```
echarrod:
```suggestion
    // Build a dynamic service, that gets the policy
```

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+            if (!validAttributes.has(key)) {
+                throw new Error(`invalid key value attribute supplied ${key}`);
+            }
+        }
+    }
+    return null;
+}
+
+/**
+ * Verify the type of remember-me.
```
echarrod: We refer to this as `RememberMeID` in the documentation and in the SDKs themselves. Looks like you call it `remember-me`/`rememberMe` throughout, if you could change the occurrences to `RememeberMeId` that would be great 👍

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+ */
+const validateRememberMe = (rememberMe) => {
+    if (typeof(rememberMe) !== "boolean") {
+        const errorMessage = `type of policy ${constants.WANTED_REMEMBER_ME} should be a boolean`;
+        throw new Error(errorMessage);
+    }
+}
+
+/**
+ * @desc validate authType is an array of boolean.
```
echarrod:
```suggestion
 * @desc validate authTypes is an array of booleans.
```

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+    if (!Array.isArray(authTypes)) {
+        const errorMessage = `type of ${constants.WANTED_AUTH_TYPES} should be an array of numbers`;
+        throw new Error(errorMessage);
+    }
+}
+
+/**
+ * 
+ * @param {*} wanted wanted is an array of attributes to be requested.
+ * @param {*} authTypes auth types represeents the authentication type to be used.
```
echarrod:
```suggestion
 * @param {*} authTypes auth types represents the authentication type to be used.
```

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+
+    policy[constants.EXTENSIONS] = extensions;
+    return policy;
+}
+
+class DynamicPolicyRequest {
+    /**
+     * 
+     * @param {*} callbackEndpoint refers to the endpoint to callback to.
+     * @param {*} policy is the poilicy to be requested, this refers to thte attributes.
```
echarrod:
```suggestion
     * @param {*} policy is the policy to be requested, this refers to the attributes.
```

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+        data[constants.EXTENSIONS] = [];
+        return data;
+    }
+};
+
+
+/**
+ * @description is the response object from connect.
+ * The response will return two values which represents the QRCode to be passed
+ * to the widget. as well as the reference id.
```
echarrod:
```suggestion
 * to the widget, and the reference ID.
```

src/yoti_common/constants.js
```diff
@@ -21,6 +21,13 @@ module.exports = Object.freeze({
   ATTR_SELFIE: 'selfie',
   ATTR_EMAIL_ADDRESS: 'email_address',
   ATTR_POSTAL_ADDRESS: 'postal_address',
-  ATTR_DOCUMENT_DETAILS: 'document_details',
```
echarrod: Was this removed accidentally?

tests/dynamic-policy.spec.js
```diff
@@ -0,0 +1,78 @@
...
+const {
+    DynamicPolicyRequest,
+    Policy,
+    getDynamicPolicy,
+} = require("../src/dynamic_policy_service");
+const Payload = require('../src/request/payload').Payload;
+
+const privateKeyFile = fs.readFileSync('./tests/sample-data/keys/node-sdk-test.pem', 'utf8');
+
+const APP_ID = "9cd51c3f-f03a-46e3-aa9b-244cbfc99d86";
```
echarrod: We don't want to check in App IDs and .pem files to the public repos, even if it's only in a test. Do you want your tests to be calling the live service? Will a mock service not suffice?

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+        data[constants.CALLBACK_ENDPOINT] = this.callbackEndpoint;
+        data[constants.POLICY] = this.policy;
+        data[constants.EXTENSIONS] = [];
+        return data;
+    }
+};
+
+
+/**
+ * @description is the response object from connect.
```
echarrod: Not sure if the same thing I mentioned above will apply here, as it's code and not documentation - referring to Connect

src/dynamic_policy_service/dynamic.policy.js
```diff
@@ -0,0 +1,165 @@
...
+ * The response will return two values which represents the QRCode to be passed
```
echarrod:
```suggestion
 * The response will return two values: the QR Code to be passed
```

README.md
```diff
@@ -22,18 +22,21 @@ How to initialise your configuration
 6) [Profile Retrieval](#profile-retrieval) -
 How to retrieve a Yoti profile using the token
 
-7) [Handling Users](#handling-users) -
+7) [Dynamic Policy Generation](#dynamic-policy-generation) -
```
qzagarese: I would title this something like `Programmatic QR code creation` as `policy` is not a concept that is exposed at an SDK level.

README.md
qzagarese:
```suggestion
How to generate a QR code link by defining the requested attributes programmatically 
```

README.md
```diff
@@ -157,6 +160,34 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
 
 [Example reference](examples/profile/index.js)
 
+
+## Dynamic Policy Generation
```
qzagarese:
```suggestion
## Programmatic QR code creation
```

index.js
```diff
@@ -3,9 +3,15 @@
 const Client = require('./src/client').YotiClient;
 const AmlAddress = require('./src/aml_type').AmlAddress;
 const AmlProfile = require('./src/aml_type').AmlProfile;
+const Policy = require('./src/dynamic_policy_service').Policy;
```
qzagarese: This variable and the corresponding module should become something like `AttributeList`

index.js
```diff
@@ -3,9 +3,15 @@
...
+const DynamicPolicyRequest = require('./src/dynamic_policy_service').DynamicPolicyRequest;
```
qzagarese: Same here with --> `DynamicAttributeListRequest`

index.js
```diff
@@ -3,9 +3,15 @@
...
+const DynamicPolicyResult = require('./src/dynamic_policy_service').DynamicPolicyResult;
```
qzagarese: Hence --> `DynamicAttributeListResult`

index.js
```diff
@@ -3,9 +3,15 @@
...
 
 module.exports = {
   Client,
   AmlAddress,
   AmlProfile,
+  Policy,
+  DynamicPolicyRequest,
```
qzagarese: Same here

index.js
```diff
@@ -3,9 +3,15 @@
...
+  DynamicPolicyResult,
```
qzagarese: Same here

src/client/index.js
```diff
@@ -3,6 +3,7 @@
 const ursa = require('ursa');
 const profileService = require('../profile_service');
 const amlService = require('../aml_service');
+const dynamicPolicyService = require('../dynamic_policy_service');
```
qzagarese: Same here

src/client/index.js
```diff
@@ -34,4 +35,19 @@ module.exports.YotiClient = class YotiClient {
...
     return amlService.performAmlCheck(amlProfile, this.pem, this.applicationId);
   }
+
+  /**
+   * @desc given a dynamic sharing request get a custom qr code denoted by the
+   * policy stated in the request.
+   * @param {*} dynamicSharingRequest is the request that contains the policy.
+   * @returns {Promise} constaining a dynamic
+   */
+  getDynamicPolicyQRCodeLink(dynamicPolicyRequest) {
```
qzagarese: this can be simply `getQRCodeLink` as `getDynamicAttributeListQRCodeLink` would be ridiculous :-D

src/client/index.js
```diff
@@ -34,4 +35,19 @@ module.exports.YotiClient = class YotiClient {
...
+    // Build a dynbanic service, that gets the service
+    return dynamicPolicyService.getDynamicPolicy(
```
qzagarese: same here --> `dynamicAttributeList`.
I'll not mention it in the remainder, but please update any reference to `policy` with `attributeList`

src/dynamic_policy_service/dynamic.policy.js
echarrod: Sorry Alttaf, I was looking at [Mike's Java PR for dynamic scenarios](https://github.com/getyoti/yoti-java-sdk/pull/58/files), and it made me realise that this `RememberMe` is with respect to the policy, not the rememberMeID of the user, apologies! So we only want to change the instances where it would be relating to the User, don't know if there are any of those.

Alttaf:
@echarrod i think all comments have been addressed. Let me know anything else you want on this, have also added pre commits to this proj

src/client/index.js
```diff
@@ -34,4 +35,19 @@ module.exports.YotiClient = class YotiClient {
...
   }
+
+  /**
+   * @desc given a dynamic attribute list request get a custom QR code denoted by the
+   * dynamic attribute list stated in the request.
+   * @param {*} dynamicAttributeListRequest is the request that contains the attribute list.
+   * @returns {Promise} containing a dynamic attribute list
+   */
+  getQRCodeLink(dynamicAttributeListRequest) {
+    // Build a dynamic service, that gets the service
```
echarrod:
```suggestion
    // Build a dynamic service, that gets the dynamic attribute list
```
`dynamic attribute list` / `QR Code`?

src/dynamic_attribute_list_service/dynamic.attribute.list.js
```diff
@@ -0,0 +1,165 @@
...
+        data[constants.EXTENSIONS] = [];
+        return data;
+    }
+};
+
+
+/**
+ * @description is the response object from Yoti.
+ * The response will return two values; the QRCode to be passed
+ * to the widget. and the reference id.
```
echarrod:
```suggestion
 * to the widget, and the reference ID.
```

tests/dynamic-attribute-list.spec.js
```diff
@@ -0,0 +1,79 @@
...
+        "optional": "false",
+    }];
+    const wantedRememberMe = false;
+    const attributeList = AttributeList(
+        wantedAttributeList,
+        null,
+        wantedRememberMe,
+    );
+    const dynamicAttributeListRequest = new DynamicAttributeListRequest(callbackURL, attributeList);
+    //const QRCODE_LINK = "https://staging0.code.yoti.com/CAEaJDRjNTQ3M2IxLTNiNzktNDg3My1iMmM4LThiMTQxZDYwMjM5ODAC";
```
echarrod: Could we remove this? We can put it in the wiki for reference.

If we want ease of use when devs are working on it, we can add it as an environment variable

README.md
```diff
@@ -24,18 +24,21 @@ How to initialise your configuration
 6) [Profile Retrieval](#profile-retrieval) -
 How to retrieve a Yoti profile using the token
 
-7) [Handling Users](#handling-users) -
+7) [Programmatic QR code creation](#dynamic-profile-attribute-requests) -
```
echarrod: Link broken after name change: 
```suggestion
7) [Programmatic QR code creation](#programmatic-qr-code-creation) -
```

davidgrayston:
@echarrod I've rebased ready for another review:
- Includes recent changes on development
- Restored a few changes that were lost in merge from `development` _(these are no longer visible in this PR)_
- Added number to commits
- Re-included examples for linting _(I'll look into how we can ensure examples are built before linting in a separate task)_
- Minor correction to example in README

src/client/index.js
davidgrayston: Now reads:
>@desc given a dynamic attribute list request get a custom QR code denoted by the dynamic attribute list stated in the request.

davidgrayston:
@echarrod in 7ade70e I've standardised the implementation inline with the Java SDK

- Introduces builders:
  - `DynamicScenarioBuilder`
  - `DynamicPolicyBuilder`
  - `ExtensionBuilder`
  - `LocationConstraintExtensionBuilder`
  - `TransactionalFlowExtensionBuilder`
- Added a simple _"Dynamic Share Example"_ to the profile example
- Updated _README_ with DynamicScenarioBuilder example

src/client/index.js
```diff
@@ -34,4 +35,20 @@ module.exports.YotiClient = class YotiClient {
   performAmlCheck(amlProfile) {
     return amlService.performAmlCheck(amlProfile, this.pem, this.applicationId);
   }
+
+  /**
+   * Given a dynamic scenario get a custom QR code denoted by the dynamic policy
```
echarrod: Not essential, but this reads a bit easier with a comma:
```suggestion
   * Given a dynamic scenario, get a custom QR code denoted by the dynamic policy
```

README.md
```diff
@@ -185,16 +185,22 @@ This QR Code/link should then be embedded into your page using the Yoti widget t
...
-        const qr-code-div = document.getElementById('qr-code-div');
-        qr-code-div.innerHTML = qrCode;
-    })
+const dynamicPolicy = new Yoti.DynamicPolicyBuilder()
+  .withFullName()
+  .build();
 
-```
+const dynamicScenario = new Yoti.DynamicScenarioBuilder()
+  .withCallbackEndpoint('/profile')
```
echarrod: I know optional arguments aren't as clean in Node as other languages, but what do you think about making this parameter optional? And having the default if it's not set be `/profile`? I'm just thinking I can't envisage clients ever wanting to change that based on the current setup, so it's just adding an extra bit of config for them

davidgrayston: I think this path/URL would vary quite often between applications and might be more risky to set a default. We'd also need to document that the default is `/profile` and changing the default in future would be a breaking change as clients would start to depend on it. (I think it's safer to require that clients specify the path for their application)

echarrod: Ok sure, fair enough

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
```diff
@@ -0,0 +1,202 @@
...
+
+    this.wantedAttributes[key] = wantedAttribute;
+    return this;
+  }
+
+  /**
+   * @param {boolean} optional
+   * @param {string} name
+   */
+  withWantedAttribute(optional, name) {
```
davidgrayston: @echarrod as discussed, I'll change this method to `withWantedAttributeByName(name, optional)` (switch order of arguments) and have `optional` default to `false`.

`addWantedAttribute()` will be renamed to `withWantedAttribute()` to match Java SDK

davidgrayston:
@echarrod please review changes in d1189d1:
- Changed `DynamicPolicyBuilder::withWantedAttribute(optional, name)` to `DynamicPolicyBuilder::withWantedAttributeByName(name, optional = false)` - this should allow for dynamically building policies from an array of attribute names without using the specific attribute methods.
- Renamed `DynamicPolicyBuilder::addWantedAttribute()` to `DynamicPolicyBuilder::withWantedAttribute()` (to match Java SDK)

src/client/index.js
davidgrayston: Fixed in afa86a6

README.md
```diff
@@ -164,6 +167,51 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
 
 [Example reference](examples/profile/index.js)
 
+
+## Programmatic QR code creation
+
+Dynamic attribute list generation is a way of requesting non-static attribute lists for your application.
```
Alttaf: Does this still make sense as below the term dynamic policy is now used?

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
```diff
@@ -0,0 +1,202 @@
...
+const SELFIE_AUTH_TYPE = 1;
+const PIN_AUTH_TYPE = 2;
+
+module.exports = class DynamicPolicyBuilder {
+  constructor() {
+    this.wantedAttributes = {};
+    this.wantedAuthTypes = [];
+  }
+
+  /**
```
Alttaf: not sue the comments add value here it would be better to add more detail about what the function is doing rather than just repeating what is in thf function params

i.e.
wantedAttribute - str, this is the wanted attribute object that is used to get the attribute by object key

Alttaf: this you could also use a tool like js docs to allow users to get generated documentation

README.md
davidgrayston: I'll change this to:
> Dynamic Policy generation is a way of requesting non-static attribute lists for your application.

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
davidgrayston: We are declaring the expected type to be passed, which should be helpful in generated docs, but I agree, it would be useful to give descriptions for the methods that aren't so obvious.

davidgrayston: Discussed this briefly with @echarrod - I've now documented the classes in 75f40bb so that it's documented in one place (rather than each method that it's passed to)

Alttaf: 👍

davidgrayston:
@echarrod please review changes in bd2cefb

- Renamed `authorisation` to `authentication` as per comments on https://github.com/getyoti/yoti-dotnet-sdk/pull/77
- Added public getters inline with other SDKs and added further test coverage for them

davidgrayston:
@echarrod please review changes in 2576de1 to remove optional arguments and methods - should be the same as https://github.com/getyoti/yoti-dotnet-sdk/pull/89

davidgrayston:
@echarrod I've added example in 2576de1...7e7296c:
- Updated browser JS so that scenario ID isn't required
- Using local Yoti package using `"yoti": "file:../.."`

davidgrayston:
@echarrod now asserting that age under/over is an integer in https://github.com/getyoti/yoti-node-sdk/pull/30/commits/3c3f03bea21b2487a738b76b287b60df949eb029

PR #33, opened 2019-01-02
SDK-501: Add comment about supported Node version

sidibos:


README.md
```diff
@@ -57,6 +57,11 @@ Yoti also allows you to enable user details verification from your mobile app by
 
 ## Requirements
 
+### Node version 
+We currently support up to Node version 8, Node version 10 is not supported.
```
echarrod: Do you think it would be useful to say why we don't support it? E.g. Node Version 10 will not be supported until [Ursa ](https://www.npmjs.com/package/ursa) is compatible with it - https://github.com/JoshKaufman/ursa/issues/167

sidibos:
I think it's better not to say, as there might be other issues related to Node 10 and we don't want to list all of them.

README.md
sidibos: I think it's better not to say, as there might be other issues related to Node 10 and we don't want to list all of them.

bucky-boy: Yep, I'd agree with @sidibos - we don't need to get into that level of detail

echarrod: Ok sure

PR #36, opened 2019-02-12
SDK-670: Send X-Yoti-SDK-Version header

davidgrayston:
- Sending `package.json` version in `X-Yoti-SDK-Version`, e.g. `3.2.0`

> I can change this to use a constant defined in `src/request/index.js` if that's preferable?

### DIFF
--- a/tests/request.spec.js
+++ b/tests/request.spec.js
@@ -47,22 +47,22 @@ describe('YotiResponse', () => {
         'X-Yoti-Auth-Key': new RegExp('^[a-zA-Z0-9]+'),
         'X-Yoti-Auth-Digest': new RegExp('^[a-zA-Z0-9]+'),
         'Content-Type': 'application/json',
-        'Accept': 'application/json',
+        Accept: 'application/json',
       };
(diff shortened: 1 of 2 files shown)


davidgrayston:
@echarrod I've added test coverage for expected headers - let me know if there's a neater/preferable way of checking these

tests/request.spec.js
```diff
@@ -37,5 +39,43 @@ describe('YotiResponse', () => {
           .catch(done);
       });
     });
+
+    context('when making an API request', () => {
+      const expectedHeaders = {
+        'X-Yoti-SDK': 'Node',
+        'X-Yoti-SDK-Version': yotiPackage.version,
+        'X-Yoti-Auth-Key': new RegExp('^[a-zA-Z0-9]+'),
```
echarrod: Since this is a public key, we can also put a length check on the RegEx.

Additionally I think we can with the digest.

davidgrayston: Updated to check length in 1a9f5626509663d674980c59ad2af2d8c169d936

PR #37, opened 2019-02-13
SDK-584: Image object

davidgrayston:
- _Abstract_ `Image` class providing methods:
  - _Abstract_ `getMimeType()`
  - `getBase64Content()`
  - `getContent()`
- `ImageJpeg`/`ImagePng` classes extend `Image`, and implement `getMimeType()`

### Backwards compatibility:
- Kept `getBase64SelfieUri()`, but is now using `Image::getBase64Content()`
- `selfie` property continues to return image content (rather than new `Image` object), but is now using `Image::getContent()`

### DIFF
--- a/src/data_type/image.js
+++ b/src/data_type/image.js
@@ -1,0 +1,40 @@
+'use strict';
+
+/**
+ * Abstract Image attribute class.
+ */
+module.exports = class Image {
+  constructor(content) {
+    if (new.target === Image) {
+      throw TypeError('new of abstract class Image');
+    }
+    this.content = content;
+  }
+
+  /**
+   * Get the raw image content.
+   *
+   * @returns {ByteBuffer}
+   */
+  getContent() {
+    return this.content;
+  }
+
+  /**
+   * Get the base64 image content.
+   *
+   * @returns {string}
+   */
+  getBase64Content() {
+    return `data:${this.getMimeType()};base64,${this.content.toBase64()}`;
+  }
+
+  /**
+   * Abstract method to return the image mime type.
+   *
+   * @returns {string}
+   */
+  getMimeType() {
+    throw new Error(`${this.constructor.name} must implement getMimeType()`);
+  }
+};
--- a/src/data_type/image.js
+++ b/src/data_type/image.js
@@ -6,7 +6,7 @@
 module.exports = class Image {
   constructor(content) {
     if (new.target === Image) {
-      throw TypeError('new of abstract class Image');
+      throw TypeError('Image is an abstract class, so cannot be instantiated');
     }
     this.content = content;
   }
--- a/src/data_type/image.js
+++ b/src/data_type/image.js
@@ -4,11 +4,21 @@
  * Abstract Image attribute class.
  */
 module.exports = class Image {
-  constructor(content) {
+  /**
+   * Image constructor.
+   *
+   * @param {ByteBuffer} content
+   * @param {string} mimeType
+   */
+  constructor(content, mimeType) {
     if (new.target === Image) {
       throw TypeError('Image is an abstract class, so cannot be instantiated');
     }
+    if (typeof mimeType === 'undefined') {
+      throw TypeError(`${this.constructor.name} must pass mimeType to the Image constructor`);
+    }
     this.content = content;
+    this.mimeType = mimeType;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/yoti_common/constants.js
```diff
@@ -23,4 +23,10 @@ module.exports = Object.freeze({
   ATTR_POSTAL_ADDRESS: 'postal_address',
   ATTR_DOCUMENT_DETAILS: 'document_details',
   ATTR_STRUCTURED_POSTAL_ADDRESS: 'structured_postal_address',
+  CONTENT_TYPE_UNDEFINED: 0,
```
echarrod: If these are only going to be used in `attribute.converter.js`, do you think we should keep them in that file? And we can move them to this file if they're going to be used in other places

tests/data-type.image.spec.js
```diff
@@ -0,0 +1,47 @@
...
+  {
+    imageObj: new ImagePng(imageContent),
+    mimeType: 'image/png',
+  },
+];
+
+imageTypes.forEach((type) => {
+  describe(`${type.imageObj.constructor.name}`, () => {
+    context('#getContent()', () => {
+      it('it should return original image content', () => {
```
echarrod: I think these tests shouldn't have the second `it`, so for this one it would just read as `it('should return original image content`

src/proto-root/proto.attribute.list.js
```diff
@@ -18,14 +20,17 @@ module.exports = {
...
+      const convertedValueByType = AttributeConverter
+        .convertValueBasedOnContentType(attrValue, attrType);
       const attrNameInCamelCase = this.toCamelCase(attrName);
 
-      attrList.push({ [attrNameInCamelCase]: convertedValueByType });
-
-      if (attrName === 'selfie') {
-        const imageUriValue = this.imageUriBasedOnContentType(attrValue, attrType);
-        attrList.push({ base64SelfieUri: imageUriValue });
+      // Handle selfies for backwards compatibility.
```
echarrod: I was originally thinking we would not update the selfie attribute to use the new image type yet, only in the major release. But if we have adequate tests to cover making sure this is backwards compatible, then it's fine to do it this way.

src/data_type/image.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+/**
+ * Abstract Image attribute class.
+ */
+module.exports = class Image {
+  constructor(content) {
+    if (new.target === Image) {
+      throw TypeError('new of abstract class Image');
```
echarrod: Could this be slightly clearer? Maybe `throw TypeError('Image is an abstract class, so cannot be instantiated');` or something similar?

davidgrayston: I've now reworded this

tests/data-type.image.spec.js
davidgrayston: I've now corrected this

src/yoti_common/constants.js
davidgrayston: I've now moved these

src/proto-root/proto.attribute.list.js
davidgrayston: I've left the API unchanged, but using the new Image object to get the existing values:
- `getBase64SelfieUri()` returns the same value as before (base64 string)
- `selfie` property continues to return the original value (not the new Image object)
-  Existing tests pass

tests/data-type.image.spec.js
echarrod: I was talking about the other `it`s in this class too, it's in a few other places 😃

davidgrayston: Now corrected the other occurrences 👍

src/data_type/image.js
```diff
@@ -0,0 +1,40 @@
...
+    return `data:${this.getMimeType()};base64,${this.content.toBase64()}`;
+  }
+
+  /**
+   * Abstract method to return the image mime type.
+   *
+   * @returns {string}
+   */
+  getMimeType() {
+    throw new Error(`${this.constructor.name} must implement getMimeType()`);
```
bucky-boy: This is fine.

But now that I look at it, I'm not sure why the Java SDK does it in the way that it does.

Rather than forcing every subclass to implement the abstract method, we could have a _protected_ constructor on the abstract base class that requires the mimeType is passed in.  Then have a _final getMimeType_ method that returns the value.

Not a big deal.  Might be slightly neater though - if it works in node.js that is.

davidgrayston: @bucky-boy I've reorganised the code in 1b4694e54a2e574e83769029a97ac22d6b24b071 - I think it's close to achieving your suggestion - does this look better?

cc @echarrod

PR #42, opened 2019-02-26
SDK-568: Deprecate getUserId()

davidgrayston:
- Introduced new `getRememberMeId()` method
- Documented activity details methods and `@deprecated` `getUserId()`
- Update example project to use `getRememberMeId()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
 
 ```
 
-Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a _Remember Me Id_.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -190,7 +190,7 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
 
 ```
 
-Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a userId.
+Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a _Remember Me Id_.
```
echarrod: Same as other comment 🙂  I**D**

davidgrayston: Now corrected

PR #43, opened 2019-02-26
SDK-568: Remove getUserId()

davidgrayston:
- Removing getUserId(), which was deprecated in #42 
- Compare with #42 https://github.com/getyoti/yoti-node-sdk/compare/SDK-568-deprecate-user-id...SDK-568-remove-user-id

⚠️ _This removal is for the `v4.x` major release_


README.md
```diff
@@ -190,7 +190,7 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
 
 ```
 
-Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a userId.
+Where `yourUserSearchFunction` is a piece of logic in your app that is supposed to find a user, given a _Remember Me Id_.
```
echarrod: If we're not doing camelcase, we should make it "Remember Me I**D**" 👍

davidgrayston: Corrected in #42 and rebased

laurent-yoti:
See #336

PR #49, opened 2019-03-05
SDK-831: Structured postal address

davidgrayston:
- Added test coverage
- Added to example project

### DIFF
--- a/examples/profile/views/pages/profile.ejs
+++ b/examples/profile/views/pages/profile.ejs
@@ -6,6 +6,7 @@ const phoneNumber = profile.getPhoneNumber();
 const dateOfBirth = profile.getDateOfBirth();
 const ageCondition = profile.getAgeVerified();
 const postalAddress = profile.getPostalAddress();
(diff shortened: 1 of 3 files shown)


echarrod:
Would you also be able to add a test which handles nested JSON for the Structured Postal Address, as this is something which could happen in the future. See Activity_AddAttributesToProfile_StructuredPostalAddress_NestedJSON in [.NET](https://github.com/getyoti/yoti-dotnet-sdk/blob/845d44ec6edb2fa13d89dcb34d7952a9685d01dc/test/Yoti.Auth.Tests/ActivityTests.cs)

davidgrayston:
> Would you also be able to add a test which handles nested JSON for the Structured Postal Address, as this is something which could happen in the future. See Activity_AddAttributesToProfile_StructuredPostalAddress_NestedJSON in [.NET](https://github.com/getyoti/yoti-dotnet-sdk/blob/845d44ec6edb2fa13d89dcb34d7952a9685d01dc/test/Yoti.Auth.Tests/ActivityTests.cs)

I've now added coverage in 65879a9

PR #50, opened 2019-03-07
SDK-759: Document Images

davidgrayston:
### MultiValue object
All `MULTI_VALUE` attribute values are converted to `MultiValue` objects:
- Allows recursive lists of any type (tree)
- Tree can be filtered by instance of or concrete type

#### Methods:
- `::getItems()` returns array of filtered items
- `::filterInstance()` allows filtering by the instance type
- `::filterType()` allows filtering by the concrete class name (`constructor.name`)
- `::applyFilters()` applies all filters
- `::resetFilter()` resets filters and values

#### Example usage:
```javascript
const multiValue = new MultiValue(items)
  .filterInstance(ImageJpeg)
  .filterType('Array')
  .filterType('MultiValue')
  .getItems();
```

### Document Images
- Added `Profile::getDocumentImages()`

  Returns an array of `Image` objects using:
  ```javascript
  multiValue
    .filterInstance(Image)
    .getItems()
  ```

### Protobuf
- Updated existing proto files
  - Syntax changed to `proto3`
  - Added `message MultiValue`, `MULTI_VALUE = 6` and `INT = 7`
  
>Note: we can add generated files in next major release


README.md
```diff
@@ -302,6 +302,19 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
   * [X] Base64 Selfie Uri `getBase64SelfieUri()`
```
echarrod: Could you please also change the `const userProfile = activityDetails.getUserProfile(); // deprecated` from the start of the API coverage section above to use `getProfile()` instead of `getUserProfile()`?

README.md
```diff
@@ -302,6 +302,19 @@ yotiClient.getActivityDetails(token).then((activityDetails) => {
...
+    * [X] Gender `getGender().getValue()`
+    * [X] Nationality `getNationality().getValue()`
+    * [X] Mobile Number `getPhoneNumber().getValue()`
+    * [X] Photo `getSelfie().getValue()`
+    * [X] Email Address `getEmailAddress().getValue()`
+    * [X] Address `getPostalAddress().getValue()`
+    * [X] Structured Address `getStructuredPostalAddress().getValue()`
+    * [X] Document Details `getDocumentDetails().getValue()`
+    * [X] Document Images `getDocumentImages().getValue()`
   * [X] userProfile `.getUserProfile()`
```
echarrod: Can we remove the userProfile section too - we don't want clients to be using it, and it makes this section very large!

src/data_type/multi.value.js
```diff
@@ -0,0 +1,83 @@
...
+    return this;
+  }
+
+  /**
+   * Apply all filters.
+   */
+  applyFilters() {
+    this.items = this.originalItems;
+
+    if (this.filterInstances.length > 0 || this.filterTypes.length > 0) {
```
echarrod: Can we split this up into `Instance` and `Type` methods? Rather than having an `or`, and doing two different things in one if statement.

davidgrayston: The reason it's currently combined is to allow either criteria being met:
```javascript
return allowedType || allowedInstance;
```

I'll see if I can achieve this in 2 methods, or at least break these into separate callbacks to make this method smaller.

README.md
davidgrayston: I've now updated this

README.md
davidgrayston: I've now removed these

src/data_type/multi.value.js
davidgrayston: @echarrod I've now split the filters into separate methods in 359dd316e08b30298c805d1de9872af1170f37dd - does this look clearer?

davidgrayston:
@echarrod please review changes in 40772cc0fe0b5a30263f387a5a38254ea4097f74
- Additional test coverage for MultiValue attributes and the sample images, which highlighted a bug in `convertMultiValue()`
- Fixed `convertMultiValue()`

davidgrayston:
@echarrod I've made a small change to ensure the document images are immutable using `Object.freeze()` in 4498836

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+  hasFilters() {
+    return this.filterInstances.length > 0 || this.filterTypes.length > 0;
+  }
+
+  /**
+   * Check instance.
+   *
+   * @param {*} value
+   */
+  allowedInstance(value) {
```
bucky-boy: isAllowedInstance?

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+module.exports = class MultiValue {
+  /**
+   * MultiValue constructor.
+   *
+   * @param {array} items
+   */
+  constructor(items) {
+    this.originalItems = items;
+    this.items = items;
+    this.filterInstances = [];
```
bucky-boy: So, this is a list of the _allowedInstances_..?

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+    this.filterTypes = [];
```
bucky-boy: And these are the _allowedTypes_...?

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  /**
+   * Filter values by their instance type.
+   *
+   * @param {*} type
+   *
+   * @returns {MultiValue}
+   */
+  filterInstance(type) {
```
bucky-boy: This adds a new _Instance type_ to the _allowedInstances_?

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Filter values by their constructor name.
+   *
+   * @param {String} type
+   *
+   * @returns {MultiValue}
+   */
+  filterType(type) {
+    this.filterTypes.push(type);
```
bucky-boy: While this adds a new _Type name_ to the list of _allowedTypes_ (or perhaps _allowedTypeNames_)

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return allowedInstance;
+  }
+
+  /**
+   * Check constructor name (class).
+   *
+   * @param {*} value
+   */
+  allowedType(value) {
```
bucky-boy: isAllowedType

src/yoti_common/attribute.converter.js
```diff
@@ -21,6 +25,13 @@ module.exports.AttributeConverter = class AttributeConverter {
     switch (attrName) {
       case constants.ATTR_DOCUMENT_DETAILS:
         return new DocumentDetails(value);
+      case constants.ATTR_DOCUMENT_IMAGES:
+        if (!(value instanceof MultiValue)) {
```
bucky-boy: I think this bit would want to fail

tests/data-type.multi.value.spec.js
```diff
@@ -0,0 +1,142 @@
...
+    });
+  });
+  context('#applyFilters', () => {
+    it('should apply filters to all multi values', () => {
+      const multiValue = new MultiValue(getTestValues());
+      expect(multiValue.getItems().length).to.equal(6);
+
+      // Apply filters.
+      multiValue
+        .filterInstance(ImageJpeg)
```
bucky-boy: This really would be easier to follow if the methods had names like _addAllowedType_, or _allowType_ or something.

What the filter really does is actually the opposite of what you'd expect from _filterType(Image)_.  I think you'd expect that to remove _Image_, rather than allow it.

src/data_type/multi.value.js
```diff
@@ -0,0 +1,110 @@
...
+    if (this.filterTypes.length > 0) {
+      allowedType = this.filterTypes.find(type => value.constructor.name === type);
+    }
+    return allowedType;
+  }
+
+  /**
+   * Reset to original values.
+   */
+  resetFilters() {
```
bucky-boy: Do we really want this?

davidgrayston:
@bucky-boy / @echarrod I've addressed points above in d6edd2d756fdc1ca551da44c17cae23a14b300ea
- Renamed `filter...()` methods to `allow...()`
- Removed `resetFilters()`
- Log `Document Images could not be decoded` if `document_images` isn't successfully converted to a `MultiValue` object (this results in a null attribute as per other failed attributes)

src/data_type/multi.value.js
davidgrayston: Removed in d6edd2d

tests/data-type.multi.value.spec.js
davidgrayston: Renamed to `allow...()` in d6edd2d

src/yoti_common/attribute.converter.js
davidgrayston: Now throwing error in d6edd2d, which will be caught and logged

davidgrayston:
Now returning `null` for:
- `MultiValue` that have any empty non-string values (as this is considered invalid)
- Empty non-string attributes

Empty strings are still allowed

Changes made in 776e195
>Note: This also covers SDK-858

PR #51, opened 2019-03-19
SDK-870: Support INT values

davidgrayston:
- Handling INT values int `AttributeConverter::convertValueBasedOnContentType()`

### DIFF
--- a/src/yoti_common/attribute.converter.js
+++ b/src/yoti_common/attribute.converter.js
@@ -15,6 +15,7 @@ const CONTENT_TYPE_DATE = 3;
 const CONTENT_TYPE_PNG = 4;
 const CONTENT_TYPE_BYTES = 5;
 const CONTENT_TYPE_MULTI_VALUE = 6;
+const CONTENT_TYPE_INT = 7;
 
 module.exports.AttributeConverter = class AttributeConverter {
   static convertValueBasedOnAttributeName(value, attrName) {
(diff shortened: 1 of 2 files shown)


src/yoti_common/attribute.converter.js
```diff
@@ -47,8 +60,25 @@ module.exports.AttributeConverter = class AttributeConverter {
         return new ImageJpeg(value);
       case CONTENT_TYPE_PNG:
         return new ImagePng(value);
+      case CONTENT_TYPE_MULTI_VALUE:
+        return AttributeConverter.convertMultiValue(value);
+      case CONTENT_TYPE_INT:
+        return parseInt(value.toUTF8(), 10);
```
echarrod: Could we add named parameters here? Just makes it clearer what the `10` is denoting 
```suggestion
        return parseInt({string: value.toUTF8(), radix: 10});
```

davidgrayston: I don't think we can do this neatly with JS. We could do one of the following:
- Assign the parameters to descriptive variables then pass those in
- Use a description constant such as `BASE_10` or `RADIX_10`
- Move `parseInt()` into a descriptive method, e.g. `parseIntFromBase10String()` that takes only the string.

It might be neater to leave this as is?

echarrod: Sure, leaving it is fine. I didn't realise the suggestion returned it as a JSON

PR #52, opened 2019-03-25
SDK-877: Travis CI

davidgrayston:
- Running tests and linting in Node 6, 8, 9, 10, 11
- Includes SDK-875 to remove `package-lock.json`

[Build Status](https://travis-ci.com/getyoti/yoti-node-sdk)


.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: node_js
+node_js:
+  - "6"
+  - "8"
+  - "9"
+  - "10"
+  - "11"
+
```
echarrod: Could we also add the following so it tests on the different OSs too:
```
os:
  - windows
  - linux
  - osx
```
detailed [here](https://blog.travis-ci.com/2018-10-11-windows-early-release)

davidgrayston: I've now added these

PR #56, opened 2019-04-09
SDK-817: Remember Me ID

davidgrayston:
- Separated `ActivityDetails` from profile service _(like other SDKs)_
- Added test coverage for `ActivityDetails` including empty _Parent Remember Me ID_ and _Remember Me ID_

>_Note: I originally added test coverage here https://github.com/getyoti/yoti-node-sdk/compare/development...SDK-817-remember-me-id, but this highlighted that ActivityDetails should be separate as it is in the other SDKs - I'll close the branch if this PR is preferable_


echarrod:
I don't think we should have `activity.details.js` under `profile_service`. ActivityDetails contains both the application and user profiles, and it has additional elements like timestamp, receiptID. Could have `activity_details_service`?

echarrod:
After discussion, happy to leave `activity_details.js` as is: based on `profile_service` being a reflection of the "/profile" endpoint it is calling

PR #60, opened 2019-04-23
SDK-923: Add getter for receipt ID

davidgrayston:
- Added getter for receipt ID

### DIFF
--- a/src/profile_service/activity.details.js
+++ b/src/profile_service/activity.details.js
@@ -99,7 +99,7 @@ module.exports.ActivityDetails = class ActivityDetails {
   }
 
   /**
-   * The Receipt ID.
+   * Receipt ID identifying a completed activity.
    *
    * @returns {string}
    */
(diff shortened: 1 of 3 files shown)


src/profile_service/activity.details.js
```diff
@@ -97,4 +97,13 @@ module.exports.ActivityDetails = class ActivityDetails {
   getBase64SelfieUri() {
     return this.profile.base64SelfieUri;
   }
+
+  /**
+   * The Receipt ID.
```
echarrod: Could we use the same description as the other SDKs ([.NET](https://github.com/getyoti/yoti-dotnet-sdk/blob/19b0bae64a96661ed4aaf0588e128c398b3e05a3/src/Yoti.Auth/ActivityDetail.cs) and [Java](https://github.com/getyoti/yoti-java-sdk/blob/f682f2ca8f07a94c0e6b4a3b68062d7be6a624a1/yoti-sdk-api/src/main/java/com/yoti/api/client/ActivityDetails.java) links). Although should we say "Receipt ID identif**ies** a completed activity." instead of "Receipt ID identifying a completed activity."?

davidgrayston: Updated in a46cc0f

PR #62, opened 2019-04-25
SDK-725: Add Dynamic Scenario example

davidgrayston:
- Added Dynamic Scenario example (moved from #30)

>Note: I'll update package.json once #30 has been released


davidgrayston:
I've moved this back to #30 as we can use `"yoti": "file:../.."` to use the local Yoti package in the example projects

PR #66, opened 2019-06-05
SDK-547: Add application profile getter to activity details

davidgrayston:
- Added application profile getter to activity details
- Existing user profile and new application profile both extend new base profile

### DIFF
--- a/src/profile_service/activity.details.js
+++ b/src/profile_service/activity.details.js
@@ -92,6 +92,8 @@ class ActivityDetails {
   /**
    * The user profile returned by Yoti if the request was successful.
    *
+   * @deprecated replaced by getProfile()
+   *
    * @returns {Object}
    */
   getUserProfile() {
@@ -99,7 +101,8 @@ class ActivityDetails {
   }
 
   /**
-   * The user Profile object.
+   * The user profile with shared attributes and anchor information, returned
+   * by Yoti if the request was successful.
    *
    * @returns {Profile}
    */
@@ -108,7 +111,7 @@ class ActivityDetails {
   }
 
   /**
-   * The application profile object.
+   * Profile of an application, with convenience methods to access well-known attributes.
    *
    * @returns {ApplicationProfile}
    */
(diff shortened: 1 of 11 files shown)


src/profile_service/activity.details.js
```diff
@@ -1,26 +1,53 @@
...
+  }
+  return profile.reduce((acc, current) => {
+    const propName = Object.getOwnPropertyNames(current)[0];
+    acc[propName] = current[propName];
+    return acc;
+  }, {});
+}
+
+/**
+ * Details of an activity between a user and the application.
```
echarrod: I was thinking this could be "an application" instead of "the application", but I see that it is like this in Java, so probably best to leave as is

src/profile_service/activity.details.js
```diff
@@ -80,6 +107,15 @@ module.exports.ActivityDetails = class ActivityDetails {
     return this.extendedProfile;
   }
 
+  /**
+   * The application profile object.
```
echarrod: Can we change this to `Profile of an application, with convenience methods to access well-known attributes.`?

Also as discussed, can we update the method description of the new profile above to reflect the fact that it is preferable, and contains anchor information too. Deprecating the old method too 👍

src/profile_service/profile.js
```diff
@@ -1,141 +1,143 @@
...
+   * Did the user pass the age verification check?
+   *
    * @returns {null|Attribute}
    */
   getAgeVerified() {
     return this.getAttribute(constants.ATTR_AGE_VERIFIED);
   }
 
   /**
+   * Corresponds to the gender in the passport; will be one of the strings "MALE", "FEMALE"
```
echarrod: I looked this up in the wiki:
```suggestion
   * Corresponds to the gender in the passport; will be one of the strings "MALE", "FEMALE", "TRANSGENDER"
```

Looks like we need to update in the other SDKs too

davidgrayston: Corrected in 09a6cc7

src/profile_service/activity.details.js
davidgrayston: Updated in 16c49a8

PR #67, opened 2019-06-07
SDK-660: Parse unknown anchors

davidgrayston:
### Added
- `Attribute.getAnchors()`
- `YotiAnchor.getType()`
- `ANCHOR_TYPES` constant to handle mapping of anchor types (`SOURCE`/`VERIFIER`/`UNKNOWN`)
- Missing method documentation
- `AnchorProcessor` now processes all anchors. Unknown anchors have empty `""` value and type `UNKNOWN`

### Deprecated
- `AnchorProcessor.getAnchorByOid()`
- `AnchorProcessor.getAnchorValueByOid()`
- `AnchorProcessor.findOidIndex()`


tests/sample-data/yoti-common/unknown-anchor.txt
```diff
@@ -0,0 +1,39 @@
+CjdBTkMtRE9Dz8qdV2DSwFJicqASUbdSRfmYOsJzswHQ4hDnfOUXtYeRlVOeQnVr3an
+                                ESmMH7e2HEqAIMIIEHDCCAoSgAwIBAgIQIrSqBBTTXWxgGf6OvVm5XDANBgkqhkiG9w0
```
echarrod: Indentation looks a bit strange here?

src/yoti_common/anchor.processor.js
```diff
@@ -266,29 +374,59 @@ module.exports.AnchorProcessor = class AnchorProcessor {
...
+   * Get anchor list key by type.
+   *
+   * @param {string} type
+   */
+  static getAnchorListKeyByType(type) {
+    const anchorTypesMap = this.getAnchorTypesMap();
+    return Object.keys(anchorTypesMap)
+      .find(key => ANCHOR_TYPES[type] === anchorTypesMap[key]);
+  }
+
```
echarrod: I was going to say that this file has gotten quite large, so it might be worthwhile to split it up. _But_ we have deprecated methods now, so should be ok once they are removed.

tests/sample-data/yoti-common/unknown-anchor.txt
davidgrayston: Fixed in 88d1edd

src/yoti_common/anchor.processor.js
davidgrayston: It also seems like some of the logic in `AnchorProcessor` could be moved to `YotiAnchor` and we may also want to rename the class to `Anchor` - could be safest to leave these as private classes for until the next major release.

davidgrayston:
@echarrod please review changes since your last review 3267efe...2c05e91:
- Split `YotiAnchor` and `YotiSignedTimeStamp` into separate files and take required arguments (no longer key/value object), which is more similar to Java implementation.
- Fixed sample data indentation.
- Added test coverage for deprecated `getAnchorByOid()` method, to ensure backwards compatibility.
- Updated anchor method documentation.

PR #75, opened 2019-07-03
Release 3.6.0: Merge development -> master

davidgrayston:
### Added
- `ActivityDetails.getApplicationProfile()`
- `ActivityDetails.getTimestamp()`
- `Attribute.getAnchors()`
- `YotiAnchor.getType()`
- `Profile.getAttributes()`
- `YotiDate` extending `Date`, providing methods:
  - `getMicroseconds()` - Number between 0 and 999999, representing the microseconds.
  - `getMicrosecondTimestamp()` - Returns ISO 8601 UTC timestamp with microseconds.
  - `getMicrosecondTime()` - Time with microseconds.
- Support for dynamic scenarios, including:
  - `YotiClient.createShareUrl()`
  - `DynamicScenarioBuilder`
  - `DynamicPolicyBuilder`
  - `WantedAttributeBuilder`
  - `ExtensionBuilder`
  - `LocationConstraintExtensionBuilder`
  - `TransactionalFlowExtensionBuilder`
- `ApplicationProfile` providing methods:
  - `getName()`
  - `getUrl()`
  - `getLogo()`
  - `getReceiptBgColor()`

### Changed
- `YotiSignedTimeStamp.getTimestamp()` now returns `YotiDate` which extends `Date`
- `AnchorProcessor` now processes all anchors. Unknown anchors have empty "" value and type `UNKNOWN`

### Deprecated
- `AnchorProcessor.getAnchorByOid()`
- `AnchorProcessor.getAnchorValueByOid()`
- `AnchorProcessor.getAnchorsByCertificate()`
- `AnchorProcessor.mergeAnchorsLists()`
- `AnchorProcessor.processSingleAnchor()`

PR #78, opened 2019-07-15
Sdk-660: Parse unknown anchors

davidgrayston:
>Correction to #67

- Now creates one `YotiAnchor` per protobuf anchor
- Reinstated `AnchorProcessor.findOidIndex()` _(Need to update #75 changelog once merged into development)_

### Deprecated
- `AnchorProcessor.getAnchorsByCertificate()`
- `AnchorProcessor.mergeAnchorsLists()`
- `AnchorProcessor.processSingleAnchor()`


src/yoti_common/anchor.processor.js
```diff
@@ -26,24 +26,77 @@ class AnchorProcessor {
...
+
+    certificatesList.forEach((certificate) => {
+      const extensionsData = AnchorProcessor.convertCertToX509(certificate).extensions;
+
+      Object.keys(ANCHOR_TYPES).forEach((type) => {
+        const oid = ANCHOR_TYPES[type];
+        const extension = this.getExtensionByOid(extensionsData, oid);
+
+        if (extension !== null) {
+          anchor = new YotiAnchor(
```
echarrod: Not sure if I'm missing something, but is there a reason why we're not just doing `return new YotiAnchor(...` here, rather than assigning that to `anchor` and then returning `anchor` below?

tests/yoti_common/anchor.processor.spec.js
```diff
@@ -119,6 +118,33 @@ describe('anchorProcessor', () => {
...
+
+      it('should return 1 verifier anchor', () => {
+        expect(anchors.verifiers).to.be.length(1);
+        expect(anchors.verifiers[0].getType()).to.equal('VERIFIER');
+      });
+
+      it('should return 1 unknown anchor', () => {
+        expect(anchors.unknown).to.be.length(1);
+        expect(anchors.unknown[0].getType()).to.equal('UNKNOWN');
+      });
```
echarrod: No harm in checking that there are 3 in total here as well is there?

davidgrayston: I've added a check in the attribute test in 5d285b0, as this test returns a key-value object, rather than a flat array of anchors.

src/yoti_common/anchor.processor.js
davidgrayston: I've changed this in 5d285b0 - I agree returning early here is clearer and inline with other SDKs

PR #81, opened 2019-07-29
SDK-1096: Fix DynamicScenarioBuilder Auth Types

MrBurtyyy:
- Update withSelfieAuthentication and withPinAuthentication to process if user passes 'false' as a parameter
- Add function to remove matching element from wantedAuthTypes (based on selfie or pin auth type)
- Filter wantedAuthTypes on export to remove duplicates.


src/dynamic_sharing_service/policy/dynamic.policy.builder.js
```diff
@@ -39,9 +46,7 @@ module.exports = class DynamicPolicyBuilder {
    * @param {string} name
    */
   withWantedAttributeByName(name) {
-    const wantedAttribute = new WantedAttributeBuilder()
-      .withName(name)
-      .build();
+    const wantedAttribute = new WantedAttributeBuilder().withName(name).build();
```
echarrod: Was this change done by the linter? It looks like we have the old line style on lines 90-92 in this file, which I prefer.

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
```diff
@@ -122,6 +127,8 @@ module.exports = class DynamicPolicyBuilder {
   withSelfieAuthentication(enabled = true) {
     if (enabled) {
       return this.withWantedAuthType(SELFIE_AUTH_TYPE);
+    } else if (!enabled) {
```
echarrod: I think we can make this more concise, since it's immediately obvious this is the path taken if `!enabled`. This will flag up on [SonarLint](https://www.sonarlint.org/) as well once we have that enabled:
```suggestion
    } else {
```

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
```diff
@@ -132,6 +139,8 @@ module.exports = class DynamicPolicyBuilder {
   withPinAuthentication(enabled = true) {
     if (enabled) {
       return this.withWantedAuthType(PIN_AUTH_TYPE);
+    } else if (!enabled) {
```
echarrod: Same:
```suggestion
    } else {
```

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
MrBurtyyy: This might have been caused by prettier which should have been following the eslint rules.  It seems that prettier is overriding some rules so I'll move it back to how it was and avoid prettier for now.

tests/dynamic_sharing_service/dynamic.scenario.builder.spec.js
```diff
@@ -1,12 +1,6 @@
...
-  expect,
-} = require('chai');
+const { expect } = require('chai');
 
-const {
-  DynamicScenarioBuilder,
-  DynamicPolicyBuilder,
-  ExtensionBuilder,
-} = require('../../');
+const { DynamicScenarioBuilder, DynamicPolicyBuilder, ExtensionBuilder } = require('../../');
```
echarrod: Same with these - if we are to change the formatting of files through an auto-formatter, I would rather we did them in one go, so as not to detract away from other PRs.

tests/dynamic_sharing_service/dynamic.scenario.builder.spec.js
```diff
@@ -81,4 +72,56 @@ describe('DynamicScenarioBuilder', () => {
...
+    expect(dynamicPolicy.wantedAuthTypes).to.not.contain(1);
+  });
+
+  it('should not have pin authentication after having it added then removed', () => {
+    const dynamicPolicy = new DynamicPolicyBuilder()
+      .withPinAuthentication(true)
+      .withPinAuthentication(false)
+      .build();
+
+    expect(dynamicPolicy.wantedAuthTypes).to.not.contain(1);
```
echarrod: Can we use the `PIN_AUTH_TYPE` and `SELFIE_AUTH_TYPE` consts here throughout the tests, instead of referring to their int values? 👍 This will be easier to manage if we change them in the future, and also makes the tests a bit clearer

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
MrBurtyyy: Eslint was complaining about this just being an "else" statement.  We might be able to remove the else completely, as due to the default parameter I think we can assume that if the original condition doesn't pass, we can attempt to remove the element from the wanted auth types.

tests/dynamic_sharing_service/dynamic.scenario.builder.spec.js
MrBurtyyy: The auth types aren't currently exported to be available to any other file.  We could move these into the class itself as static variables, but checking the Java SDK these are private as well.  I think it would be good to give visibility to the rest of the application for these, but it's a decision that needs to be made.

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
MrBurtyyy: [This commit](https://github.com/getyoti/yoti-node-sdk/commit/067bdfe0f960725d452e43c112f4adbf47f3c8c1) should fix this.  Tests are still passing

tests/dynamic_sharing_service/dynamic.scenario.builder.spec.js
MrBurtyyy: [Fixed here.](https://github.com/getyoti/yoti-node-sdk/commit/067bdfe0f960725d452e43c112f4adbf47f3c8c1) . I have disabled prettier until further notice

tests/dynamic_sharing_service/dynamic.scenario.builder.spec.js
MrBurtyyy: We could move these constants into the constructor for the class therefore making them accessible to other components:

`
this.SELFIE_AUTH_TYPE = 1;
this.PIN_AUTH_TYPE = 2;
`

I tried using a 'get' function but eslint didn't like the fact that it wasn't using 'this' inside of a class method.

Currently the module only exports the class so without changing how the module export works I think this is the best solution (as the constants are directly related to the policy builder)

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
echarrod: Looks good 👍

tests/dynamic_sharing_service/dynamic.scenario.builder.spec.js
MrBurtyyy: After some more looking at this, I think the constants would probably be better off as static 'get' functions within the DynamicPolicy class (rather than the builder which should focus solely on building an actual DynamicPolicy).  Example below:
`
get SELFIE_AUTH_TYPE() {
  return 1;
}

// Call with the following
DynamicPolicy.SELFIE_AUTH_TYPE
`

Thoughts @echarrod ?

echarrod: Seems ok to me, but I think we can wait for @davidgrayston to get back tomorrow to get his thoughts 👌

tests/dynamic_sharing_service/policy/dynamic.policy.builder.spec.js
```diff
@@ -210,7 +210,7 @@ describe('DynamicPolicyBuilder', () => {
       .withWantedAuthType(99)
       .build();
 
-    const expectedAuthTypes = [1, 2, 99];
```
davidgrayston: As discussed, we should try to avoid reusing constants in tests - we can redefine them in the test to make the test easier to follow

src/dynamic_sharing_service/policy/dynamic.policy.js
```diff
@@ -33,6 +33,20 @@ module.exports = class DynamicPolicy {
     this.wantedRememberMe = wantedRememberMe;
   }
 
+  /**
+   * Constant for selfie authorisation type.
+   */
+  static get SELFIE_AUTH_TYPE() {
```
davidgrayston: As discussed, we can remove these if we redefine in the test

tests/dynamic_sharing_service/policy/dynamic.policy.builder.spec.js
```diff
@@ -273,4 +273,56 @@ describe('DynamicPolicyBuilder', () => {
...
+
+  it('should build with no more than one auth type', () => {
+    const dynamicPolicy = new DynamicPolicyBuilder()
+      .withSelfieAuthentication(true)
+      .withSelfieAuthentication()
+      .withSelfieAuthentication(true)
+      .build();
+
+    const authTypesLength = dynamicPolicy.getWantedAuthTypes().length;
+    expect(authTypesLength).to.equal(DynamicPolicy.SELFIE_AUTH_TYPE);
```
davidgrayston: This should remain `2`, as we are checking the length of auth types

tests/dynamic_sharing_service/policy/dynamic.policy.builder.spec.js
```diff
@@ -273,4 +273,56 @@ describe('DynamicPolicyBuilder', () => {
...
+  });
+
+  it('should build with only two auth types', () => {
+    const dynamicPolicy = new DynamicPolicyBuilder()
+      .withSelfieAuthentication(true)
+      .withPinAuthentication(true)
+      .build();
+
+    const authTypesLength = dynamicPolicy.getWantedAuthTypes().length;
+    expect(authTypesLength).to.equal(DynamicPolicy.PIN_AUTH_TYPE);
```
davidgrayston: Same as above - we should check the length is 2

PR #85, opened 2019-08-14
SDK-1125: Dynamic Scenario - Source Constraints

davidgrayston:
### Added
- `WantedAnchorBuilder`
  - `.withValue()`
  - `.withSubType()`
  - `.build()`
- `ConstraintsBuilder`
  - `.withSourceConstraint()`
  - `.build()`
- `SourceConstraintBuilder`
  - `.withAnchor()`
  - `.withAnchorByValue()`
  - `.withSoftPreference()`
  - `.withPassport()`
  - `.withDrivingLicence()`
  - `.withNationalId()`
  - `.withPasscard()`
  - `.build()`
- `WantedAttributeBuilder.withConstraints()`
- `WantedAttributeBuilder.withAcceptSelfAsserted()`
- `DynamicPolicyBuilder.withDocumentImages()`
- `DynamicPolicyBuilder.withDocumentDetails()`

You can try out fetching passport or driving license by updating the dynamic scenario example:
```javascript
const dynamicPolicy = new Yoti.DynamicPolicyBuilder()
  .withFullName()
  .withSelfie()
  .withDocumentImages(constraints)
  .withDocumentDetails(constraints)
  .build();
```


README.md
```diff
@@ -184,14 +184,24 @@ This QR Code/link should then be embedded into your page using the Yoti widget t
 The following example demonstrates how a Dynamic Policy can be built using attribute methods such as `withFullName()`, and generic methods `withWantedAttribute()` and `withWantedAttributeByName()`.
 
 ```javascript
+const sourceConstraint = new yoti.SourceConstraintBuilder()
```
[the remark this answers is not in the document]
davidgrayston: Addressed in f24cbafb6fb057c5255a11023ea70df9f2242988

README.md
```diff
@@ -184,14 +184,24 @@ This QR Code/link should then be embedded into your page using the Yoti widget t
...
-  .withName('email_address')
+  .withName('given_names')
+  .withConstraints(constraints)
+  .withAcceptSelfAsserted()
   .build();
 
 const dynamicPolicy = new yoti.DynamicPolicyBuilder()
   .withFullName()
-  .withWantedAttributeByName('given_names')
+  .withWantedAttributeByName('email_address')
```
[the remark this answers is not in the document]
davidgrayston: Addressed in f24cbafb6fb057c5255a11023ea70df9f2242988

src/dynamic_sharing_service/policy/wanted.attribute.js
```diff
@@ -38,14 +51,42 @@ module.exports = class WantedAttribute {
...
+   * List of constraints.
+   *
+   * @returns {Constraints}
+   */
+  getConstraints() {
+    return this.constraints;
+  }
+
+  /**
+   * Accept self asserted.
```
echarrod:
```suggestion
   * Accept self asserted attributes. These are attributes that have been self-declared, and not verified by Yoti.
```

src/dynamic_sharing_service/policy/wanted.attribute.js
```diff
@@ -38,14 +51,42 @@ module.exports = class WantedAttribute {
     return this.derivation;
   }
 
+  /**
+   * List of constraints.
```
echarrod:
```suggestion
   * List of constraints to add to an attribute. If you do not provide any particular constraints Yoti will provide you with the information from the most recently added source.
```

davidgrayston: Updated in 5fda572

src/dynamic_sharing_service/policy/wanted.attribute.js
davidgrayston: Updated in 5fda572

src/dynamic_sharing_service/policy/source.constraint.builder.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const SourceConstraint = require('./source.constraint');
+const WantedAnchorBuilder = require('./wanted.anchor.builder');
+
+const ANCHOR_PASSPORT = 'PASSPORT';
```
MrBurtyyy: I think I commented on the Python PR the same thing, but I think these should live in the same place as where other constants are kept (i.e. where we keep the attribute constants).  That way if we need access somewhere else in the SDK, we don't have to maintain it in two places

davidgrayston: I believe I discussed this with @echarrod - we have the helper methods for these document types, so there's not really a need to expose the constants at the moment.

However, it might help someone using the `withAnchorByValue()` method - @echarrod what do you think?

We'd probably want to make them more descriptive if we do this, somethinng like`ANCHOR_VALUE_PASSPORT`

echarrod: Would there be a reason that they'd use `withAnchorByValue("Passport")`, rather than the helper method?

If not, I think we can keep them here, and then we can always move them to the common constants file if we need them elsewhere (assuming they're not exported here).

Also renaming to `ANCHOR_VALUE_PASSPORT` seems sensible, even if we don't move it

davidgrayston: Renamed in c32df35ea3b3922699d23a984bc526b4ae33eddc

PR #86, opened 2019-08-15
SDK-1134: Wanted Attribute Constraints

davidgrayston:
> On hold until #85 is complete and once this feature is supported. See https://github.com/getyoti/yoti-node-sdk/compare/SDK-1125-source-constraints...SDK-1134-wanted-attribute-constraints for changes in that PR.

### Added
- `BaseProfile.getAllAttributesByName()`
- `proto.attribute.list.decodeAllAttributeListData()`

### Changed
- We can now request the same attribute with different constraints


src/profile_service/index.js
```diff
@@ -17,7 +17,7 @@ module.exports.getReceipt = (token, pem, appId) => {
         try {
           const receipt = response.getReceipt();
           const parsedResponse = response.getParsedResponse();
-          const decryptedProfile = yotiCommon.decryptCurrentUserReceipt(receipt, pem);
+          const decryptedProfile = yotiCommon.decryptCurrentUserReceipt(receipt, pem, true);
```
MrBurtyyy: Why do we need this boolean?  Shouldn't we always just parse the full list of attributes?

davidgrayston: I agree, this wasn't the best approach - I've refactored in ecf8863 and updated the PR description, so it should be a little better now.  This needs to be refactored in the next major, which we have a ticket for.

src/dynamic_sharing_service/policy/dynamic.policy.builder.js
```diff
@@ -30,8 +32,17 @@ module.exports = class DynamicPolicyBuilder {
     if (wantedAttribute.getDerivation()) {
       key = wantedAttribute.getDerivation();
     }
+    if (wantedAttribute.getConstraints()) {
+      const hash = crypto
+        .createHash('sha256')
+        .update(JSON.stringify(wantedAttribute.getConstraints()), 'utf8')
+        .digest('hex');
+
+      key = `${key}:${hash}`;
```
MrBurtyyy: Super pedantic, but is it worth keeping this the same as the other SDK's where I think we use `-` instead of `:`?

davidgrayston: Changed in d1ec770

davidgrayston:
Now processing `attributesMap` once in 7307585 to prevent filtering each time an attribute getter is called (re-requesting review)

davidgrayston:
@echarrod using selfie image data URI in 997963c

laurent-yoti:
Closing as merged in #311

PR #87, opened 2019-08-16
SDK-578: Add support for multiple age verifications

davidgrayston:
### Added
- `AgeVerification`
  - `.getCheckType()`
  - `.getAge()`
  - `.getResult()`
  - `.getAttribute()`
- `BaseProfile`
  - `.findAttributesStartingWith()`
- `Profile`
  - `.getAgeVerifications()`
  - `.findAgeOverVerification(age)`
  - `.findAgeUnderVerification(age)`
  - `.findAgeVerification(type, age)` - intended to be used by _findAgeOverVerification_/_findAgeUnderVerification_
  - `.findAllAgeVerifications()`
- `Validation.matchesPattern()`

### Deprecated
- `Profile.getAgeVerified()`
- `Age` found at _yoti_common/age_

### DIFF
--- a/examples/profile/views/pages/profile.ejs
+++ b/examples/profile/views/pages/profile.ejs
@@ -4,7 +4,6 @@ const givenNames = profile.getGivenNames();
 const familyName = profile.getFamilyName();
 const phoneNumber = profile.getPhoneNumber();
 const dateOfBirth = profile.getDateOfBirth();
-const ageCondition = profile.getAgeVerified();
 const postalAddress = profile.getPostalAddress();
 const emailAddress = profile.getEmailAddress();
 const nationality = profile.getNationality();
@@ -40,12 +39,6 @@ const attributes = [
     icon: "yoti-icon-calendar"
   },
   {
-    name: "Age verified",
-    preValue: "Age Verification/",
-    prop: ageCondition,
-    icon: "yoti-icon-verified"
-  },
-  {
     name: "Address",
     prop: postalAddress,
     icon: "yoti-icon-address"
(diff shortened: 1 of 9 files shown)


examples/profile/views/pages/profile.ejs
```diff
@@ -39,12 +38,6 @@ const attributes = [
     prop: dateOfBirth,
     icon: "yoti-icon-calendar"
   },
-  {
-    name: "Age verified",
-    preValue: "Age Verification/",
```
echarrod: I think we can get rid of preValue from the example project altogether now, since it was only being used for this.

Would you be able to update the sdk-static-pages repo to remove this, and also to add the AgeVerifications display in? 👍

src/data_type/age.verification.js
```diff
@@ -0,0 +1,66 @@
...
+
+    const split = attribute.getName().split(':');
+    this.checkType = split[0];
+
+    this.age = parseInt(split[1], 10);
+    this.result = attribute.getValue() === 'true';
+  }
+
+  /**
+   * The type of age check performed, as specified on Yoti Hub. Currently this might
```
echarrod: I'm assuming you might have got this from another SDK, but I'm not a fan of using the word "currently", as it'll quickly get outdated! I'm more of a fan of taking the approach we do with sources i.e. `among the possible options are age_under and age_over`

davidgrayston: Updated in 22a7aec

examples/profile/views/pages/profile.ejs
davidgrayston: Updated in 22a7aec

davidgrayston:
@echarrod I've replaced Object.values for Node 6 compatibility in 2a1a22a

PR #90, opened 2019-09-11
Release 3.7.0

davidgrayston:
- [x] Requires #94

#### Signed Request Builder
- `RequestBuilder`
- `YotiRequest`
- `YotiResponse` _(now exported)_
- `buildConnectApiRequest()` (in _./request/index.js_) for use by connect API services.
#### Age Verifications
- `AgeVerification`
  - `.getCheckType()`
  - `.getAge()`
  - `.getResult()`
  - `.getAttribute()`
- `BaseProfile`
  - `.findAttributesStartingWith()`
- `Profile`
  - `.getAgeVerifications()`
  - `.findAgeOverVerification(age)`
  - `.findAgeUnderVerification(age)`
  - `.findAgeVerification(type, age)`
  - `.findAllAgeVerifications()`
#### Document Details
- `DocumentDetails.getIssuingAuthority()`

### Deprecated
- `makeRequest()` (in _./request/index.js_)
- `Profile.getAgeVerified()`
- `Age` found at _yoti_common/age_

### Fixed
- Updated example project to run in Node 6


echarrod:
Change notes are nice and detailed for us here, but I think we can condense for the actual release.

E.g. "Dynamic Scenario Source Constraints" is not going to be used by many people, so I don't think we need the whole section under that. 

And also I'm unsure if we need to include `Profile.getPostalAddress() will now return formatted address from structured address attribute when postal address is null.`, as this process should be invisible to our clients. I can't remember what we've done in other SDKs for this though...

Will `Validation.matchesPattern()` be used by clients? If not, we can remove from change notes

davidgrayston:
I've checked that `acceptSelfAsserted` works as expected

PR #100, opened 2019-09-27
SDK-604: Sandbox

davidgrayston:
### Added
- `SandboxClientBuilder` builds `SandboxClient`
- `SandboxAttributeBuilder` builds `SandboxAttribute`
- `SandboxAgeVerificationBuilder` builds `SandboxAgeVerification`
- `SandboxAnchorBuilder` builds `SandboxAnchor`
- `TokenRequestBuilder` builds `TokenRequest`
- `YotiDate.fromDateString()`
- `YotiDate.toISODateString()`

### Fixed
- Exception message tests now check messages are correct (regression when switching to Jest)


src/sandbox/client.js
```diff
@@ -0,0 +1,62 @@
...
+        })
+        .catch((err) => {
+          console.log(`Error retrieving requested data: ${err}`);
+          return reject(err);
+        });
+    });
+  }
+}
+
+module.exports = {
```
MrBurtyyy: Are these better off being exported as `module.exports = SandboxClient`?  Generally I think you only return an object if you are exporting more than one class/component as you are just creating an object and then instantly destructuring it when you import

src/sandbox/index.js
```diff
@@ -0,0 +1,13 @@
+const { SandboxClientBuilder } = require('./client.builder');
```
MrBurtyyy: Further to my previous comment, these would become for example: `const SandboxClient = require('./client.builder');`

src/sandbox/profile/request/attribute/anchor.builder.js
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * @class SandboxAnchorBuilder
+ */
+class SandboxAnchorBuilder {
+  /**
+   * Setup initial property values.
+   */
+  constructor() {
+    this.anchors = [];
```
MrBurtyyy: Whats this used for?

src/sandbox/profile/request/attribute/anchor.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * @class SandboxAnchor
+ */
+class SandboxAnchor {
+  /**
+   * @param {string} type
+   * @param {string} value
+   * @param {*} subType
```
MrBurtyyy: Isn't sub-type always a string?

src/sandbox/profile/request/attribute/attribute.js
```diff
@@ -0,0 +1,94 @@
...
+   * @returns {boolean}
+   */
+  getOptional() {
+    return this.optional;
+  }
+
+  /**
+   * @returns {SandboxAnchor[]}
+   */
+  getAnchors() {
```
MrBurtyyy: Does Node not implement the same helper methods as Python and Java to easily filter the sources and verifiers?  Don't think it's needed, but would be good to keep them all similar?

src/sandbox/profile/request/token.builder.js
```diff
@@ -0,0 +1,249 @@
...
+    return this.withAttribute(sandboxAttribute);
+  }
+
+  /**
+   * @returns {TokenRequest}
+   */
+  build() {
+    return new TokenRequest(
+      this.rememberMeId,
+      Object.keys(this.attributes).map(k => this.attributes[k])
```
MrBurtyyy: Would it not be better to store the attributes in an array originally?  By using a keyed dictionary, aren't we limiting the sandbox token to only allow one attribute with a certain name? (Also we wouldn't have to convert from a dictionary to array when building the request)

src/sandbox/client.js
davidgrayston: Fixed in 6c963d8

src/sandbox/index.js
davidgrayston: Fixed in 6c963d8

src/sandbox/profile/request/attribute/anchor.builder.js
davidgrayston: Removed in e47de2c

src/sandbox/profile/request/attribute/attribute.js
davidgrayston: _Java SDK_ has only `getAnchors()` - I think this is ok for sandbox testing purposes

src/sandbox/profile/request/token.builder.js
davidgrayston: _Java SDK_ uses a map similar to this - as discussed we can change to array/list when we add support for multiple attributes with the same name, but different constraints

src/sandbox/profile/request/attribute/anchor.js
davidgrayston: Corrected in e6aa460

davidgrayston:
I've moved sandbox out of `src` in e6aa460...4077687

Sandbox can now be required as follows:

```javascript
const {	
  SandboxClientBuilder,	
  SandboxAttributeBuilder,	
  SandboxAgeVerificationBuilder,	
  SandboxAnchorBuilder,	
  TokenRequestBuilder,	
} = require('yoti/sandbox');
```

PR #103, opened 2019-10-03
SDK-1208: Request Builder Headers

davidgrayston:
### Changed
- `X-Yoti-Auth-Key` header is only provided to profile endpoint
- Default `Content-Type` header is only added when payload is provided

### DIFF
--- a/tests/request/request.builder.spec.js
+++ b/tests/request/request.builder.spec.js
@@ -1,54 +1,22 @@
 const nock = require('nock');
 const fs = require('fs');
 
-const { YotiRequest } = require('../../src/request/request');
-const { RequestBuilder } = require('../../');
+const { RequestBuilder, Payload } = require('../../');
 const yotiPackage = require('../../package.json');
 
 const PEM_FILE_PATH = './tests/sample-data/keys/node-sdk-test.pem';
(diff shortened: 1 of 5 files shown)


tests/request/request.builder.spec.js
```diff
@@ -1,54 +1,22 @@
...
-      done();
-    })
-    .catch(done);
+const CONTENT_TYPE_HEADER_NAME = 'Content-Type';
+const CONTENT_TYPE_JSON = 'application/json';
+const DEFAULT_HEADERS = {
+  'X-Yoti-SDK': 'Node',
+  'X-Yoti-SDK-Version': `Node-${yotiPackage.version}`,
+  Accept: CONTENT_TYPE_JSON,
+  'X-Yoti-Auth-Digest': new RegExp('^[a-zA-Z0-9/+=]{344}$'),
```
echarrod: Can we easily use the `DIGEST_KEY_PATTERN = /^[a-zA-Z0-9/+=]{344}$/` value here, instead of redeclaring? No worries if not

davidgrayston: We could move all common test constants/data in to a separate file, but we'd want to do this across all tests. As discussed, we'll leave it as is for now.

PR #104, opened 2019-10-08
SDK-1236: Supported Node version

davidgrayston:
### Changed
- Compatible Node engine in _package.json_ to `>=6`


davidgrayston:
@echarrod if we put a specific number, it would probably need to be `>=6.17.1` (which currently runs on Travis), `>=6` behaved the same as `>=6.0.0` and seemed neater

PR #107, opened 2019-10-21
SDK-1255: Fix YotiDate format

MrBurtyyy:
## Changed

* Updated `YotiDate` to return microsecond time format with minutes zero padded

### DIFF
--- a/src/data_type/date.js
+++ b/src/data_type/date.js
@@ -58,7 +58,7 @@ class YotiDate extends Date {
    */
   getMicrosecondTime() {
     const hours = formatDatePart(this.getUTCHours(), 2);
-    const minutes = formatDatePart(this.getUTCMinutes());
+    const minutes = formatDatePart(this.getUTCMinutes(), 2);
     const secondsMicroseconds = formatSecondsWithMicroseconds(
       this.getUTCSeconds(),
       this.getMicroseconds()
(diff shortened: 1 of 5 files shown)


src/data_type/date.js
```diff
@@ -58,7 +58,7 @@ class YotiDate extends Date {
    */
   getMicrosecondTime() {
     const hours = formatDatePart(this.getUTCHours(), 2);
-    const minutes = formatDatePart(this.getUTCMinutes());
+    const minutes = formatDatePart(this.getUTCMinutes(), 2);
     const secondsMicroseconds = formatSecondsWithMicroseconds(
       this.getUTCSeconds(),
       this.getMicroseconds()
```
echarrod: Do we need to add formatting and tests for seconds and microseconds too?

MrBurtyyy: I can turn this test into one that will cover all of these.  The test would then be something like `should return zero padded string for hours, minutes and seconds`

MrBurtyyy: Fixed in https://github.com/getyoti/yoti-node-sdk/pull/107/commits/72f4c05f9ff036e21dd279e68576f62ba371d334

MrBurtyyy:
> What's the merge plan to get these changes into Development?

We'll merge this into master, then open another PR from this branch into development (following the Gitflow workflow) @emmas-yoti

PR #110, opened 2019-10-31
SDK-1262: Third Party Attribute Extension

MrBurtyyy:
- [x] Merge #108 into `development`

## Added

* `ThirdPartyAttributeExtensionBuilder`
* `ThirdPartyAttributeExtensionContent`

See [here](https://github.com/getyoti/yoti-node-sdk/compare/SDK-1231-CredentialIssuanceDetails...SDK-1262-ThirdPartyAttributeExtension) for comparison as this is branched off SDK-1231


src/dynamic_sharing_service/extension/third.party.attribute.extension.content.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+const Validation = require('../../yoti_common/validation');
+
+class ThirdPartyAttributeExtensionContent {
+  constructor(expiryDate, definitions) {
+    Validation.instanceOf(expiryDate, Date, 'expiryDate');
+    Validation.hasOnlyStringValues(definitions, 'definitions');
```
davidgrayston: Perhaps this could be an array or `Definition`, that will serialise to `{ "name": "some name"  }` using `.toJSON()`.

This would prevent a future breaking change if we need to add a new _definition_ property.

src/dynamic_sharing_service/extension/third.party.attribute.extension.content.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  getDefinitions() {
+    return this.definitions;
+  }
+
+  toJSON() {
+    return {
+      expiry_date: this.expiryDate.toISOString(),
+      definitions: this.definitions.map(i => ({ name: i })),
```
davidgrayston: If this is an array of `Definition` that serialise to JSON, we wont need to map here.

src/dynamic_sharing_service/attribute.definition.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  toJSON() {
+    return {
+      name: this.name,
+    };
+  }
+}
+
+module.exports = AttributeDefinition;
```
davidgrayston: Should this file be in `src/data_type/` if it's to be reused for #110 ?

src/dynamic_sharing_service/extension/third.party.attribute.extension.builder.js
```diff
@@ -0,0 +1,42 @@
...
+    this.definitions = [];
+  }
+
+  withExpiryDate(expiryDate) {
+    Validation.instanceOf(expiryDate, Date, 'expiryDate');
+    this.expiryDate = expiryDate;
+    return this;
+  }
+
+  withDefinition(definition) {
```
davidgrayston: As discussed, we should pass a string to the method used to create an `AttributeDefinition`

MrBurtyyy: Fixed in https://github.com/getyoti/yoti-node-sdk/pull/110/commits/8b8c3f343791dacba873ac9d5eab36cd76575698

src/dynamic_sharing_service/attribute.definition.js
MrBurtyyy: Fixed in https://github.com/getyoti/yoti-node-sdk/pull/110/commits/8b8c3f343791dacba873ac9d5eab36cd76575698

davidgrayston:
I've exported `ThirdPartyAttributeExtensionBuilder` in 0b96475 so that it's available as:

```node
const { ThirdPartyAttributeExtensionBuilder } = require('yoti');
```

PR #127, opened 2020-02-12
Release 3.8.1

davidgrayston:
### Fixed
- Document details parsing: now accommodates more values _(including redacted Aadhar numbers)_
- `YotiDate.fromDateString()` now supports microsecond accuracy
- `YotiDate.constructor()` now supports signed microsecond timestamps, e.g. `-1571630945999999` will have **1** microsecond
- Rounding when converting microseconds to `YotiDate`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "yoti",
-  "version": "3.8.0",
+  "version": "3.8.1",
   "description": "Yoti NodeJS SDK for back-end integration",
   "author": "Yoti LTD <tech@yoti.com> (https://www.yoti.com/developers)",
   "license": "MIT",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "yoti",
-  "version": "3.8.0",
+  "version": "3.8.1",
```
emmas-yoti: Not related to this PR but going forwards we're updating version numbers to the candidate number after a release, not before it.

echarrod: I was originally think we should do this for the benefit of sonar scanning. I don't think it's as important to update it here after each release, but I guess we might as well if we're updating it for sonar.

PR #140, opened 2020-02-27
NA: Enable engine-strict on Travis

davidgrayston:
### Changed
- Enabled `engine-strict` to fail when installing incompatible packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,9 @@ jobs:
   include:
     - stage: Test
       node_js: "10"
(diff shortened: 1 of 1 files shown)


davidgrayston:
Closing this PR as Travis isn't failing with `engine-strict` enabled.

Tried the following:

#### Setting config
```
npm config set engine-strict true
```

#### Setting variable in command
```
npm_config_engine_strict=true npm install
```

#### Setting env variable
```
env:	
  - npm_config_engine_strict=true
```

PR #142, opened 2020-03-05
SDK-1301: Doc Scan Demo

davidgrayston:
### Added
- Doc Scan Demo

> #### Notes
> - Uses as much OOTB [Bootstrap](https://getbootstrap.com/docs/4.4/getting-started/introduction/) as possible
> - Copy `.env.example` to `.env` and configure stage URLs


examples/doc-scan/.env.example
```diff
@@ -0,0 +1,8 @@
+# Required configuration.
+# The client SDK ID is generated by https://hub.yoti.com when you publish your app.
```
echarrod:
```suggestion
# The client SDK ID and PEM file for your application are generated on https://hub.yoti.com.
```

davidgrayston:
@echarrod It seemed to make sense to split all example READMEs now, all with a link from root README 885c496

See how it looks here: https://github.com/getyoti/yoti-node-sdk/blob/SDK-1301-doc-scan-client-demo/README.md#running-the-examples

PR #161, opened 2020-04-15
IN-2431: update Yoti share script and allow css spacing for inline QR…

nikhilPank:
… codes


echarrod:
Closing, addressing in new PR

PR #208, opened 2020-07-06
SDK-1637: Doc Scan Chip Data

davidgrayston:
### Added
- Doc Scan chip data support

### Fixed
- Downgraded `eslint-plugin-jest` for Node 6


src/doc_scan_service/session/create/task/requested.text.extraction.config.js
```diff
@@ -11,10 +11,15 @@ class RequestedTextExtractionConfig {
...
    *   Describes the manual fallback behaviour applied to each Task
+   * @param {string} chipData
+   *   Describes the chip data requirement for each Task
    */
-  constructor(manualCheck) {
+  constructor(manualCheck, chipData) {
     Validation.isString(manualCheck, 'manualCheck');
     this.manualCheck = manualCheck;
+
+    Validation.isString(chipData, 'chipData', true);
```
echarrod, line 21: Will this break anyone's code who is currently calling this with just manual check? In .NET this is public and so I was going to default it to empty string

davidgrayston: `true` is the optional flag, so it can be `string` or `undefined` - It will be excluded from the payload by default

davidgrayston: https://github.com/getyoti/yoti-node-sdk/blob/master/src/yoti_common/validation.js#L17

echarrod: Cool, sounds good 👍

PR #241, opened 2020-09-29
SDK-1703: Supplementary Documents

davidgrayston:
### Added
- Supplementary document support

### Deprecated
- `TaskResponse.getGeneratedTextDataChecks()` this method is now implemented on subclasses


examples/doc-scan/src/controllers/index.controller.js
```diff
@@ -46,10 +50,15 @@ async function createSession() {
     )
     .withRequestedTask(
       new RequestedTextExtractionTaskBuilder()
-        .withManualCheckNever()
+        .withManualCheckAlways()
```
echarrod, line 53: Do we want to change this?

src/doc_scan_service/session/create/filters/required.supplementary.document.builder.js
```diff
@@ -0,0 +1,53 @@
...
+const Validation = require('../../../../yoti_common/validation');
+const Objective = require('./objective/objective');
+
+class RequiredSupplementaryDocumentBuilder {
+  /**
+   * @param {string[]} countryCodes
+   *
+   * @returns {this}
+   */
+  withCountryCodes(countryCodes) {
```
echarrod, line 13: Do we have the singular WithCountryCode & WithDocumentType for DocumentFilters? If so, I think we should add here too

examples/doc-scan/src/controllers/index.controller.js
davidgrayston: This was intentional to help demonstrate checks in the response

src/doc_scan_service/session/create/filters/required.supplementary.document.builder.js
davidgrayston: This hasn't been added to Java, so perhaps we should leave this out for now?

echarrod (approved):
👍

PR #259, opened 2021-01-14
SDK-1828 Third party identity check

taranchauhan:
## Added

Third party identity check

- [X] Added support for the check using the request builder
- [X] Implement retrieving the check from a session
- [x] Add check to the Doc scan example when creating and retrieving the session


vishmimoney (approved):

laurent-yoti (approved):
Look good!

PR #260, opened 2021-01-21
SDK-1857 migrate to GitHub actions

laurent-yoti:
### Description

Migrating the tests run in Travis to Github Actions (work from @emmas-yoti)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -14,6 +14,10 @@ jobs:
 
       - uses: actions/setup-node@v2.1.2
 
+      - run: npm install
+
+      - run: npm update
+
       - run: npm audit --production
       
       - run: set -o pipefail && npm audit --audit-level=high --parseable | awk -F $'\t' 'NF {print $2,"("$3") - "$4" - "$5" - "$6}' | awk '!visited[$0]++'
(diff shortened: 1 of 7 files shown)


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,77 @@
+name: Tests
+on: [push, pull_request_target]
+
+jobs:
+  tests:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    if: >
+      github.event_name == 'push' ||
```
taranchauhan, line 9: Anyway to move this into a common variable or something as I saw the same condition in all the if's.

Not really sure if possible with Github actions. Only saw the option to define something at the workflow level as an environment variable.

taranchauhan (approved):

.github/workflows/tests.yaml
laurent-yoti: Do you mean the `push` and `pull` to be constant define somewhere ?
If so, I would say that this would probably have the inconvenient of not being the same as on the other repos... and also that I am not sure that this may have been generated like that by Github itself when you use the UI to create a workflow

vishmimoney (approved):
Looks good.

PR #277, opened 2021-06-28
SDK-1836 New auth-type (BEARER)

laurent-yoti:
### Description

Allows the NotificationConfig to have a authorisation type of either `BASIC` or `BEARER`.
The NotificationConfigBuilder exposes two methods `withAuthTypeBasic()` and `withAuthTypeBearer()`

### Testing

All CI tests pass

### DIFF
--- a/src/doc_scan_service/session/create/notification.config.js
+++ b/src/doc_scan_service/session/create/notification.config.js
@@ -1,13 +1,16 @@
 'use strict';
 
+const DocScanConstants = require('../../doc.scan.constants');
 const Validation = require('../../../yoti_common/validation');
+
+const acceptedAuthTypes = [DocScanConstants.BASIC, DocScanConstants.BEARER];
 
 /**
  * Configures call-back Notifications to some backend endpoint provided
  * by the Relying Business.
  *
  * Notifications can be configured to notified a client backend of certain
- * events, avoiding the neeed to poll for the state of the Session.
+ * events, avoiding the need to poll for the state of the Session.
  *
  * @class NotificationConfig
  */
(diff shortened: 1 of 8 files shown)


src/doc_scan_service/session/create/notification.config.js
```diff
@@ -1,29 +1,40 @@
...
    *   The authorization token to be included in call-back messages
+   * @param {string} authType
+   *   The authorization type to used in call-back messages, accepts BASIC, BEARER
    * @param {string} endpoint
    *   The endpoint that notifications should be sent to
    * @param {string[]} topics
    *   The list of topics that should trigger notifications
    */
-  constructor(authToken, endpoint, topics) {
+  constructor(authToken, authType, endpoint, topics) {
```
LuborRod, line 28: I know no-one should be calling this constructor, since we want them to use the builder. But if they are, will the addition of an argument (which isn't the last one) cause existing calls to this method to use the wrong parameters?

laurent-yoti: Yes, this is a breaking change indeed. 
At first, I thought about introducing it in last position, but I changed opinion: 
- as mentioned, the builder should be the main/only recommended way to interact with that constructor
- using the [Java as reference](https://github.com/getyoti/yoti-java-sdk/blob/13d14b9e3ec180021dce7556f5b976d3058ddbdd/yoti-sdk-api/src/main/java/com/yoti/api/client/docs/session/create/NotificationConfig.java#L29), I thought implementation consistency was more desirable

LuborRod: @echarrod  What do you think about it?

echarrod: RE: 2nd point - we noticed this after it had been released in Java, and so [opted to put it in the last position for PHP](https://github.com/getyoti/yoti-php-sdk/pull/229#discussion_r652579477), with a view to doing the same in the other SDKs. 

Whilst they _shouldn't_ be using this constructor, my view was that we might as well make it backwards compatible, since it is a small change

laurent-yoti: I am fine either way.
The other argument for keeping the breaking change is that it makes the class constructor signature a bit cleaner: the 2 auth params are grouped, as opposed to being the first and last params, and given we are unlikely to revisit this code later, the full argument is to say that it is better to have it now looking good, rather than never.

Anyway, I will make the change, so it will be consistent with PHP.

echarrod (approved):
:+1:

PR #295, opened 2021-12-16
SDK-1983 Add support for session deadline

michaelyoti:
* Added support for the session_deadline property in the session specification
* Added a test case for the new property

### DIFF
--- a/src/doc_scan_service/session/create/session.specification.js
+++ b/src/doc_scan_service/session/create/session.specification.js
@@ -37,7 +37,6 @@ class SessionSpecification {
    */
   constructor(
     clientSessionTokenTtl,
-    sessionDeadline,
     resourcesTtl,
     userTrackingId,
     notifications,
@@ -45,7 +44,8 @@ class SessionSpecification {
     requestedTasks,
     sdkConfig,
     requiredDocuments,
-    blockBiometricConsent
+    blockBiometricConsent,
+    sessionDeadline
   ) {
     Validation.isInteger(clientSessionTokenTtl, 'clientSessionTokenTtl', true);
     this.clientSessionTokenTtl = clientSessionTokenTtl;
--- a/src/doc_scan_service/session/create/session.specification.js
+++ b/src/doc_scan_service/session/create/session.specification.js
@@ -16,6 +16,8 @@ class SessionSpecification {
   /**
    * @param {int} clientSessionTokenTtl
    *   Client-session-token time-to-live to apply to the created session
+   * @param {Date} sessionDeadline
+   *   The deadline that the session needs to be completed by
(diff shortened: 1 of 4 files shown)


src/doc_scan_service/session/create/session.specification.js
```diff
@@ -35,6 +37,7 @@ class SessionSpecification {
    */
   constructor(
     clientSessionTokenTtl,
+    sessionDeadline,
```
laurent-yoti, line 40: It seems logical to have it as second parameter, but if we were to consider anyone using this constructor directly, that would be a breaking change.
Any reason (beyond the grouping by theme) to have it in second position, was it like that in the Java SDK ?

src/doc_scan_service/session/create/session.specification.js
```diff
@@ -47,6 +50,11 @@ class SessionSpecification {
     Validation.isInteger(clientSessionTokenTtl, 'clientSessionTokenTtl', true);
     this.clientSessionTokenTtl = clientSessionTokenTtl;
 
+    if (sessionDeadline) {
```
laurent-yoti, line 53: Just wondering about the handling of `clientSessionTokenTtl` and `sessionDeadline` given they are mutually excluding, the API would reject, but I am considering maybe getting that handle by the SDK too, what do you think

src/doc_scan_service/session/create/session.specification.js
michaelyoti: Ah yes, you're right. I didn't consider that. The reason for putting it there was just logical grouping, but if it's going to be breaking then I don't think it's worth it

src/doc_scan_service/session/create/session.specification.js
michaelyoti: I was considering that too, but this comment says not handle on the SDK: https://lampkicking.atlassian.net/browse/SDK-1977?focusedCommentId=196347

laurent-yoti: Ok, so from the comment, I understand that we would let the API handles the case when both are present, fine.
Then, I am a bit annoyed that the code is not 100% mirroring the handling of `clientSessionTokenTtl`, aka, that there is this check `if (sessionDeadline)`... but I appreciate that this is because the `Validation.instanceOf` does not have the `optional` parameter like `Validation.isInteger`. So, all good.

src/doc_scan_service/session/create/session.specification.js
laurent-yoti: Yeah, let's change it in case that is introducing a breaking change for some users (despite they should be using the builder).

laurent-yoti (approved):
Looks good, please merge

PR #298, opened 2022-01-13
SDK-1890 add attempts configuration

laurent-yoti:
Added 3 methods in **SdkConfigBuilder** to allow setting up the retries of ID Document text extraction by category.
```
withIdDocumentTextExtractionCategoryRetries(category, retries)
withIdDocumentTextExtractionReclassificationRetries(retries)
withIdDocumentTextExtractionGenericRetries(retries)
```
The values are expressed in the sdk config within the `attempts_configuration` field:
```
// sdk config
{
  [...]
  attempts_configuration: {
    ID_DOCUMENT_TEXT_DATA_EXTRACTION: {
      GENERIC: 3,
      RECLASSIFICATION: 2,
    },
  },
}
```

### DIFF
--- a/src/doc_scan_service/session/create/sdk.config.builder.js
+++ b/src/doc_scan_service/session/create/sdk.config.builder.js
@@ -31,7 +31,7 @@ class SdkConfigBuilder {
   /**
    * Sets the allowed capture method
    *
-   * @param {string} allowedCaptureMethod the allowed capture method
+   * @param {string} allowedCaptureMethods the allowed capture method
    *
    * @returns {this}
    */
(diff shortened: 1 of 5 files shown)


src/doc_scan_service/session/create/sdk.config.builder.js
```diff
@@ -165,6 +165,61 @@ class SdkConfigBuilder {
...
+   * @param {number} retries the number of retries (more than 0)
+   *
+   * @returns {this}
+   */
+  withIdDocumentTextExtractionCategoryRetries(category, retries) {
+    Validation.isString(category, 'category');
+    Validation.isInteger(retries, 'retries');
+    Validation.notLessThan(retries, 1, 'retries');
+
+    const attemptsTarget = DocScanConstants.ID_DOCUMENT_TEXT_DATA_EXTRACTION;
```
michaelyoti, line 181: Is there a plan to have different attempts targets in the future?

laurent-yoti: Hum, not that I can see in the documentation we have, but it could happen.
Would that affect the implementation ? I suppose the creation of the overall `attemptsConfiguration` map could then be generalise, so you'd have a getOrCreate() internal method somewhere so it is not repeated in every different targets method.

michaelyoti (approved):

PR #305, opened 2022-03-24
SDK-2074 support identity profile requirements in dynamic policy

laurent-yoti:
Updated dynamic.policy and its builder to expose `withIdentityProfileRequirements()`, which then set `identityProfileRequirements` in the policy.

~(includes #304, branch to be rebased once it is merged)~ now rebased

### DIFF
--- a/src/dynamic_sharing_service/policy/dynamic.policy.js
+++ b/src/dynamic_sharing_service/policy/dynamic.policy.js
@@ -19,7 +19,7 @@ module.exports = class DynamicPolicy {
     wantedAttributes,
     wantedAuthTypes,
     wantedRememberMe = false,
-    identityProfileRequirements
+    identityProfileRequirements = null
   ) {
     Validation.isArrayOfType(wantedAttributes, WantedAttribute, 'wantedAttribute');
     this.wantedAttributes = wantedAttributes;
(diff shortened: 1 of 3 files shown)


manuel-yoti (approved):

src/dynamic_sharing_service/policy/dynamic.policy.js
```diff
@@ -13,11 +13,13 @@ module.exports = class DynamicPolicy {
...
    * @param {integer[]} wantedAuthTypes - auth types represents the authentication type to be used.
    * @param {boolean} wantedRememberMe
+   * @param {object} identityProfileRequirements
    */
   constructor(
     wantedAttributes,
     wantedAuthTypes,
-    wantedRememberMe = false
+    wantedRememberMe = false,
+    identityProfileRequirements
```
michaelyoti, line 22: Minor point, but having a required parameter after an optional parameter makes the default value of the optional parameter redundant

laurent-yoti: I see you point, agreeing with the incorrect feeling it provides.
Yet, that parameter `identityProfileRequirements` is not mandatory, it is just new (and therefore we place it last), so should it be written as `identityProfileRequirements = undefined` in your opinion ?

michaelyoti: Oh I see. Yes, I prefer to be explicit about whether it's required or not. If it's not required I prefer `= null`.

michaelyoti (approved):

PR #309, opened 2022-04-01
SDK-2106 Introduce identity_profile to session retrieve

liamd-yoti:


src/doc_scan_service/session/retrieve/identity.profile.response.js
```diff
@@ -0,0 +1,51 @@
...
+
+class IdentityProfileResponse {
+  constructor(identityProfile) {
+    Validation.isString(identityProfile.subject_id, 'subject_id', true);
+    this.subjectId = identityProfile.subject_id;
+
+    Validation.isString(identityProfile.result, 'result');
+    this.result = identityProfile.result;
+
+    Validation.isPlainObject(identityProfile.failure_reason, 'failure_reason');
```
manuel-yoti, line 14: Aren't `failure_reason` and `identity_profile_report` optional ?

In this corresponding .NET ticket, it's mentioned that these 2 are optional (_and only present in certain cases_).

In any case, not sure if we need much/any of such validation. This data isn't coming from an external client, but from out own Yoti services. So basically such validation tests the behaviour of our Yoti services. So is this ok ?

Just as an example, would it be ok to implement more in-depth validation logic, such as: `failure_reason` = present only if result="ABORDED" ? Such approach might be bad, since we'd duplicate the internal logic performed inside Doc Scan API, and we'd then have to keep it in sync inside the SDKs.

So how much (if any) response validation should we perform inside the SDKs ?

src/doc_scan_service/session/retrieve/identity.profile.report.schemes.compliance.response.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const Validation = require('../../../yoti_common/validation');
+
+class identityProfileReportSchemesComplianceResponse {
```
laurent-yoti, line 5: With capital first letter!

src/doc_scan_service/session/retrieve/identity.profile.report.schemes.compliance.response.js
```diff
@@ -0,0 +1,39 @@
...
+  constructor(schemesCompliance) {
+    Validation.isPlainObject(schemesCompliance.scheme, 'scheme');
+    this.scheme = schemesCompliance.scheme;
+
+    Validation.isBoolean(schemesCompliance.requirements_met, 'requirements_met');
+    this.requirementsMet = schemesCompliance.requirements_met;
+
+    Validation.isString(schemesCompliance.requirements_not_met_info, 'requirements_not_met_info');
+    this.requirementsNotMetInfo = schemesCompliance.requirements_not_met_info;
```
laurent-yoti, line 14: It seems this one should be optional, cause in case the requirements are met, then there is not much to say, right ?

src/doc_scan_service/session/retrieve/identity.profile.response.js
laurent-yoti: > Aren't failure_reason and identity_profile_report optional ?

Yes, they are, in the sense these fields could simply not exist or have nil values (depending on how the session flow happened).

Typically, in the SDK, we are just using a pattern of nil check before in such cases, hence here something like that would do:
```
if (identityProfile.identity_profile_report) { // can add the plain object check as well after 
  this.identityProfileReport = new IdentityProfileReportResponse(...)
}
```

Regarding the validation, as much as it seems a bit useless, we'll keep it consistent to the rest of the SDK code.

As per the `failure_reason`, the .NET has been handling it with an extra class... which is arguable... but for the sake of being aligned, we should do the same, please see the [`FailureReasonResponse`](https://github.com/getyoti/yoti-dotnet-sdk/pull/365/files)... maybe should be named though a bit more specifically, `IdentityProfileFailureReasonResponse` ?

liamd-yoti: Sure so now those properties (failure_reason and identity_profile_report) are optional, and failure_reason now has its own class as suggested.

src/doc_scan_service/session/retrieve/identity.profile.report.schemes.compliance.response.js
liamd-yoti: Oops! Corrected.

src/doc_scan_service/session/retrieve/identity.profile.report.schemes.compliance.response.js
liamd-yoti: yep, updated it to be optional.

laurent-yoti (approved):
Happy with that!

PR #310, opened 2022-04-01
SDK-2102 Introduce new identity profile to session specification.

liamd-yoti:
### Usage

```js
const identityProfileRequirements = {
  trust_framework: 'UK_TFIDA',
  scheme: {
    type: 'DBS',
    objective: 'STANDARD'
  }
}

const sessionSpec = new SessionSpecificationBuilder()
  .withIdentityProfileRequirements(identityProfileRequirements)
  .build()
```


tests/doc_scan_service/session/create/session.specification.builder.test.js
```diff
@@ -159,4 +175,46 @@ describe('SessionSpecificationBuilder', () => {
...
+      requested_tasks: [],
+      required_documents: [],
+    });
+
+    expect(JSON.stringify(sessionSpec)).toBe(expectedJson);
+  });
+
+  it('should build SessionSpecification with identityProfile', () => {
+    const identityProfileRequirementsDescriptor = {
+      trust_framework: 'UK_TFIDA',
```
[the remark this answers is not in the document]
manuel-yoti: This comment is very minor, I approved the PR anyway 👍

michaelyoti (approved):

tests/doc_scan_service/session/create/session.specification.builder.test.js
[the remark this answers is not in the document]
liamd-yoti: Just to be clear, it is not any type that can be passed to `withIdentityProfile()`, but an object. So while you could pass any set of keys and values, I think the test illustrates that an object must be passed. I think the documentation would be more responsible for providing the exact shape supported by the `identity_profile_requirements`.

src/doc_scan_service/session/create/session.specification.builder.js
```diff
@@ -154,6 +154,12 @@ class SessionSpecificationBuilder {
     return this;
   }
 
+  withIdentityProfile(identityProfile) {
```
laurent-yoti, line 157: Sorry for being pedantic and annoying, I checked also on the .NET version, and this should be called `withIdentityProfileRequirements`, which then means that `identityProfile` should become `identityProfileRequirements`...

src/doc_scan_service/session/create/session.specification.builder.js
```diff
@@ -154,6 +154,12 @@ class SessionSpecificationBuilder {
...
+    Validation.isPlainObject(identityProfile, 'identityProfileConfig');
```
laurent-yoti, line 158: Should be 
```
Validation.isPlainObject(identityProfileRequirements, 'identityProfileRequirements')
```
The first param is what you want to validate, the second one is the name of the fields so if an error is raised, it will refer to that field.

tests/doc_scan_service/session/create/session.specification.builder.test.js
```diff
@@ -159,4 +175,46 @@ describe('SessionSpecificationBuilder', () => {
...
+      .build();
+
+    const expectedJson = JSON.stringify({
+      requested_checks: [],
+      requested_tasks: [],
+      required_documents: [],
+    });
+
+    expect(JSON.stringify(sessionSpec)).toBe(expectedJson);
+  });
```
laurent-yoti, line 190: I see the intention, but I don't think that this is valuable to add.
As a matter of fact, previous test are already asserting that when you do not call `.withIdentityProfile` (soon to be renamed), then the payload does not include it.
So I think we can remove this test.

tests/doc_scan_service/session/create/session.specification.builder.test.js
```diff
@@ -159,4 +175,33 @@ describe('SessionSpecificationBuilder', () => {
 
     expect(JSON.stringify(sessionSpec)).toBe(expectedJson);
   });
+
+  it('should build SessionSpecification with identityProfile', () => {
```
laurent-yoti, line 179: Ahah, missed one `identityProfile` here ;-)

laurent-yoti (approved):
Looks good, one little last remaining `identityProfile`.
Then good to merge.

PR #312, opened 2022-04-04
SDK-2110 Introduced subject to IDV request

liamd-yoti:
Usage:
```js
...
const subject = {
  subject_id: 'some_subject_id_string'
}

const sessionSpec = new SessionSpecificationBuilder()
  .withIdentityProfile(someIdentityProfile)
  .withSubject(subject)
  .build();
```


manuel-yoti (approved):

tests/doc_scan_service/session/create/session.specification.builder.test.js
```diff
@@ -64,6 +64,10 @@ describe('SessionSpecificationBuilder', () => {
       .withFilter(documentFilter)
       .build();
 
+    const subjectDescriptor = {
```
laurent-yoti, line 67: Your test is absolutely valid, yet, I would prefer to extract the `withSubject` to its own separated test.
That is cause `withSubject` is not meant to be used outside of the TFIP context (as a matter of fact, the IDV API could throw an error, at best ignore it).
This test file initially was 1 test that was adding all the options in one session builder call, but, then we introduce more granular tests as we added options.

src/doc_scan_service/session/create/session.specification.builder.js
```diff
@@ -154,6 +154,12 @@ class SessionSpecificationBuilder {
     return this;
   }
 
+  withSubject(subject) {
+    Validation.isPlainObject(subject, 'subject');
```
michaelyoti, line 158: Is there no validation on the object keys/values?

manuel-yoti: That's done downstream, by DocScan API. 

We don't want to duplicate their logic, or take on their responsibility

michaelyoti: Why do we do any validation then? When we implement Typescript in the SDK, we'll have to be explicit about what the object's properties are. In that case developers using the SDK will have compile-time validation (and auto-complete for all config inputs), and so we could ditch any runtime validation done here and just leave that for docscan API.

Regardless, this seems to be the pattern for now, so it's fine.

michaelyoti (approved):

src/doc_scan_service/session/create/session.specification.builder.js
liamd-yoti: That is a good point, as typescript is really gonna move us towards being specific with the typing (keys and values), so either we just continue this approach of "loose" validation (only the top level type of what is being passed) or we become more granular (property names and value types).

manuel-yoti: Small note: the addition of Typescript (TS) on top of this codebase does not make it mandatory to type the entire config data. **As a first step to such TS integration**, we would probably keep the same level of validation as we do now, without going into nested data shape.

If we ditch the run-time validation, would that mean that non-TS integrators would have no validation ??
If so, something like zod might make a lot of sense.

As a 2nd step to such TS integration, we can add more input config validation. For instance by relying on the Doc Scan API OpenAPI schema, so that we avoid duplication the logic in here.

michaelyoti: Yes, vanilla JS developers would not get the validation. I am in favour of having comprehensive validation at the SDK level regardless of TS/JS, so +1 for zod

src/doc_scan_service/session/create/session.specification.js
```diff
@@ -76,6 +79,11 @@ class SessionSpecification {
     if (identityProfileRequirements) {
       Validation.isPlainObject(identityProfileRequirements, 'identityProfileRequirements');
       this.identityProfileRequirements = identityProfileRequirements;
+
+      if (subject) {
+        Validation.isPlainObject(subject, 'subject');
+        this.subject = subject;
+      }
```
laurent-yoti, line 86: Yeah, i get the intent, I did not do that on the Digital-ID side as I left it up to the IDV API to reject.
So, please also extract that subject out of the `identityProfileRequirements` even if it is true that specifying a subject outside of the context of the TFIP would not work.

tests/doc_scan_service/session/create/session.specification.builder.test.js
```diff
@@ -204,4 +212,59 @@ describe('SessionSpecificationBuilder', () => {
...
+      .build();
+
+    const expectedJson = JSON.stringify({
+      requested_checks: [],
+      requested_tasks: [],
+      required_documents: [],
+    });
+
+    expect(JSON.stringify(sessionSpec)).toBe(expectedJson);
+  });
```
laurent-yoti, line 269: Same here related to my previous comment....
We do not care here about validating that the subject is passed in the context of the `identityProfileRequirements`.

I know it sounds annoying and bizarre, but such was the decision on the others SDK... and for flexibility reason, trying to keep it loose.

laurent-yoti (approved):
Please merge!

PR #313, opened 2022-04-05
[CI] Exclude 'node-forge' from audit

laurent-yoti:
Changed the actions "Unit Tests" to use 'audit-ci' so 'node-forge' can be bypassed.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,11 +17,13 @@ jobs:
 
       - uses: actions/setup-node@v2.1.2
 
-      - run: npm install
-
       - run: npm update
 
-      - run: npm audit --production --audit-level=moderate
+      - run: npm install --no-audit
+
+      - run: npm install -D audit-ci
+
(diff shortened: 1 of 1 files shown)


liamd-yoti (approved):

GuillaumeCisco:
Why does node-forge need to be excluded from audit?

laurent-yoti:
@GuillaumeCisco 
> Why does node-forge need to be excluded from audit?

It is excluded as the project is using an old version of `node-forge`, which gets flagged by `npm audit`, still better to ignore it than not running any audit at all.
We ought to address our dependencies update at some point, but that will be looked at in the scope of a major upgrade (when we will stop support for Node 6, hopefully soon)

GuillaumeCisco:
Thank you @laurent-yoti 
This is very important to us, `node-forge` concerns all our critical issues. And we cannot get rid of it until `yoti` is up to date.
I think this should be flagged as high priority.
If your manager or your boss disagree, I can talk to him directly.

PR #314, opened 2022-04-05
SDK-2118 Add support for CIFAS ID check on session retrieval

liamd-yoti:
Covers check type of `THIRD_PARTY_IDENTITY_FRAUD_1`


src/doc_scan_service/doc.scan.constants.js
```diff
@@ -13,6 +13,7 @@ module.exports = Object.freeze({
   ID_DOCUMENT_FACE_MATCH: 'ID_DOCUMENT_FACE_MATCH',
   ID_DOCUMENT_COMPARISON: 'ID_DOCUMENT_COMPARISON',
   THIRD_PARTY_IDENTITY: 'THIRD_PARTY_IDENTITY',
+  THIRD_PARTY_IDENTITY_FRAUD_1: 'THIRD_PARTY_IDENTITY_FRAUD_1',
```
manuel-yoti, line 16: I think the value of this constant should be `CIFAS_IDENTITY`.

This corresponds to the `type` of a `check`, as defined in the linked ticket, belonging to the Doc Scan API team: https://lampkicking.atlassian.net/browse/DOCS-1554.

PS: THIRD_PARTY_IDENTITY_FRAUD`_1` would anyway be a pretty bad name IMO, as it would be very confusing, and not descriptive when compared to the normal `THIRD_PARTY_IDENTITY_FRAUD`, or other potential future THIRD_PARTY_IDENTITY_FRAUD`_n`

MrBurtyyy: @manuel-yoti @liamd-yoti This won't be `CIFAS_IDENTITY`.  Those tickets were written before product asked us to change it, and we haven't yet written tickets for how this will be exposed on the public API.

liamd-yoti: I see, think we may have jumped on this one a bit early then. Thanks for clearing that up 👍

manuel-yoti: Thanks for pointing this out @MrBurtyyy 
Can we continue this thread in a Jira ticket ? 

I left a comment about this topic in here: https://lampkicking.atlassian.net/browse/SDK-2118

liamd-yoti: So it has been settled to be: `THIRD_PARTY_IDENTITY_FRAUD_1`, will revert.

manuel-yoti (approved):

michaelyoti (approved):

src/doc_scan_service/session/retrieve/third.party.identity.fraud.1.check.response.js
```diff
@@ -0,0 +1,21 @@
+const GeneratedProfileResponse = require('./generated.profile.response');
+const CheckResponse = require('./check.response');
+
+class ThirdPartyIdentityFraud1CheckResponse extends CheckResponse {
```
laurent-yoti, line 4: Please have a look at `ThirdPartyIdentityCheckResponse` as it is doing the same thing, except it actually gets that for free as it extends the `ProfileCheckResponse`... so we should do the same for this one.

tests/doc_scan_service/session/retrieve/third.party.identity.fraud.1.check.response.spec.js
```diff
@@ -0,0 +1,88 @@
+const ThirdPartyIdentityFraud1CheckResponse = require('../../../../src/doc_scan_service/session/retrieve/third.party.identity.fraud.1.check.response');
+const GeneratedProfileResponse = require('../../../../src/doc_scan_service/session/retrieve/generated.profile.response');
+const GeneratedMedia = require('../../../../src/doc_scan_service/session/retrieve/generated.media');
+const ReportResponse = require('../../../../src/doc_scan_service/session/retrieve/report.response');
+
+describe('ThirdPartyIdentityFraud1CheckResponse', () => {
```
laurent-yoti, line 6: Following my comment about this `ThirdPartyIdentityFraud1CheckResponse` class could extend `ProfileCheckResponse`, and given that the `ProfileCheckResponse` test are already doing what is done within this test file, I think we could simply assert the here that the inheritance is in place.

liamd-yoti: I've removed all of the tests related to the `GeneratedProfileResponse` and also the ones inherited from `CheckResponse`. There is a single test that is asserting that the `.getGeneratedProfile` method returns an instance of `GeneratedProfileResponse`. Not sure if the intention is to keep this?

laurent-yoti (approved):
Please merge

PR #319, opened 2022-04-11
SDK-2091 get attribute by id

laurent-yoti:
Updated the Attribute so it assigns a field `id` if present within raw attribute data argument.
Added in BaseProfile the method getAttributeById()
Amended proto.attribute.list so the ephemeralId field from attribute proto is read and assign to the raw attribute id field.

### DIFF
--- a/tests/profile_service/profile.spec.js
+++ b/tests/profile_service/profile.spec.js
@@ -71,7 +71,6 @@ const profileDataAsArrayWithSameAttributes = [...loadProfileDataArray(), ...load
     }
     return attr;
   });
-console.log('>>>>', profileDataAsArrayWithSameAttributes, '<<<< profileDataAsArrayWithSameAttributes');
 
 describe('Profile', () => {
   [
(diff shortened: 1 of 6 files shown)


michaelyoti (approved):

tests/profile_service/profile.spec.js
```diff
@@ -63,16 +63,25 @@ const createAttribute = (name, value) => new Attribute({
   verifiers: [],
 });
 
+const profileDataAsObject = loadProfileData();
+const profileDataAsArrayWithSameAttributes = [...loadProfileDataArray(), ...loadProfileDataArray()]
+  .map((attr, index) => {
+    if (['selfie', 'document_images'].includes(attr.name)) {
+      return Object.assign({}, attr, { id: `${attr.name}-${index}` });
```
michaelyoti, line 70: Why not use destructuring?

laurent-yoti: This is all over the project, this is not using it due to the Node v6 support.

Once we move to 12 minimum, we'll be chasing the code to remove Object.assign.

PR #320, opened 2022-05-15
SDK-2142 Surface error details in share receipt response

laurent-yoti:
Modified in profile_service activity.details.js, so it exposes a getErrorDetails() method, that returns an object corresponding to the 'error_details' from the response.

```javascript
// given a response with failure
{
  "session_data": { ... },
  "receipt": {
     sharing_outcome: "FAILURE"
   },
  "error_details": { 
    "error_code": "the-error-code-from-the-failure-receipt"
    "description": "some error desc"
  }
}

// usage:
const activityDetails = await getReceipt(token, pem, sdkId)
const outcome = activityDetails.getOutcome()
console.log(outcome) // FAILURE
const errorDetails = activityDetails.getErrorDetails()
console.log(errorDetails)
// {
//   errorCode: "the-error-code-from-the-failure-receipt"
//   description: "some error desc"
// }
```

### DIFF
--- a/src/profile_service/activity.details.js
+++ b/src/profile_service/activity.details.js
@@ -146,8 +146,8 @@ class ActivityDetails {
     let errorDetails;
     if (responseErrorDetails) {
       errorDetails = {
-        errorCode: errorDetails.error_code,
-        description: errorDetails.description,
+        errorCode: responseErrorDetails.error_code,
+        description: responseErrorDetails.description,
       };
     }
     return errorDetails;
--- a/src/profile_service/activity.details.js
+++ b/src/profile_service/activity.details.js
@@ -142,14 +142,15 @@ class ActivityDetails {
 
   getErrorDetails() {
     // eslint-disable-next-line camelcase
-    const errorDetails = this.parsedResponse.error_details;
-    if (errorDetails) {
-      return {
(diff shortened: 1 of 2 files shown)


src/profile_service/activity.details.js
```diff
@@ -140,6 +140,18 @@ class ActivityDetails {
...
+  getErrorDetails() {
+    // eslint-disable-next-line camelcase
+    const errorDetails = this.parsedResponse.error_details;
+    if (errorDetails) {
+      return {
+        errorCode: errorDetails.error_code,
+        description: errorDetails.description,
+      };
+    }
+    return undefined;
```
michaelyoti, line 152: This is redundant

michaelyoti (approved):

src/profile_service/activity.details.js
laurent-yoti: Yup, sure, but that makes `eslint` happy... hence present.
Other way would be to write a ternary....

laurent-yoti: or not making the return in the `if` statement, declaring a variable above and returning it no matter what after the `if`...

laurent-yoti: Made the change anyway...

PR #326, opened 2022-06-13
SDK-2148 Clients and services to accept custom API url

laurent-yoti:
Updated services (`doc_scan_service`, `aml_service`, `dynamic_sharing_service` and `profile_service`) to:
- expose a class that encapsulates `sdkId`, `pem` and `options` (apiUrl)
- still expose their static method

Updated clients (`YotiClient`, `DocScanClient`) classes to accept options (apiUrl), and therefore have its services configured with it. This makes both clients a bit more aligned.

Deleted `/request/index` as it is only relevant to 'Connect' related calls.


src/client/yoti.client.js
```diff
@@ -29,13 +29,23 @@ class YotiClient {
...
 
+    const options = {
+      apiUrl: apiUrl || config.yoti.connectApi,
+    };
+
+    this.amlService = new AmlService(sdkId, pem, options);
+    this.profileService = new ProfileService(sdkId, pem, options);
+    this.dynamicShareService = new DynamicShareService(sdkId, pem, options);
+
     /** @deprecated replaced by this.sdkId */
```
liamd-yoti, line 47: Can this `@deprecated` be removed and the line that follows too?

src/client/yoti.client.js
```diff
@@ -29,13 +29,23 @@ class YotiClient {
...
+   * @param {Object} options
+   * @param {string} options.apiUrl
    */
-  constructor(sdkId, pem) {
+  constructor(sdkId, pem, { apiUrl } = {}) {
     this.sdkId = sdkId;
     this.pem = pem;
 
+    const options = {
+      apiUrl: apiUrl || config.yoti.connectApi,
```
liamd-yoti, line 40: Minor: I see that the apiUrl here is being decided on via some logic inside of the `options`. For consistency should we keep that logic inside of the constructor destructuring like how it is used in other places:

aml_service usage: `{ apiUrl = DEFAULT_API_URL } = {}`

liamd-yoti:
Looks good overall 👍

michaelyoti:
I'm not sure I understand the reason behind converting these to classes? They made more sense as functions IMO. I guess the validation is not repeated but that's pretty lightweight.

src/client/yoti.client.js
laurent-yoti: Well, this probably been there for some time...
I would like to remove it indeed, but we shall do that in as part of a major release with breaking changes, wouldn't it be better?

src/client/yoti.client.js
laurent-yoti: Yes, good point!

The rational for having a different way of doing is not very strong, mainly about saying that `services` options are directly consumed and store within the class instances, therefore must be passed as valid or defaulted in case of.
On the other hand, the `clients` are instantiating their services and "need" to assemble an options object (which in the case of the `YotiClient` is shared), so it seemed more appropriate to clearly state that preparation phase as a separate block (which you can also then share across the different service) and also could allow more logic to be done when defaulting some values...
As mentioned, not very strong argument... but I guess I went this way as in the end it helps dissociating the `clients` and `services` too.

laurent-yoti:
> I'm not sure I understand the reason behind converting these to classes? They made more sense as functions IMO. I guess the validation is not repeated but that's pretty lightweight.

The use of class here is "justified" by:
- trying to align `clients` and `services` definitions... (which is not quite critical to do, agreed), and `services` amongst themselves (`DocScanService` was already a class)
- allowing `services` to use a different base url on each call while reducing their methods complexity

So only point 2 really was the starting point.

Prior to change, we had the following (except for `DocScanService`)
```javascript
// serviceX

export function abc(parameter, pem, sdkId)

// where `parameter` is really related to what the function `abc` needs to do, while the `pem` and `sdk` are "options" for the eventual call.
```

The goal is now to be able to add a `apiUrl`
A approach would have been:
```javascript
// serviceX

export function abc(parameter, pem, sdkId, apiUrl)

// cons: ever growing list of side parameters..., having to implement that in many places, not consistent with DocScanService, order of parameters...

// or

export function abc(parameter, {pem, sdkId, apiUrl})
// cons: better but still loads of refactoring
```

The other approach was to say that a service as such can be a instance of a class, aka, a preconfigured bunch of method, which allow multiple variations of configuration without leaving a trail in the code (aka reducing arguments passing).

Happy to read about some approach idea.

michaelyoti:
@laurent-yoti  Thanks for the explanation. I think creating a class to reduce the complexity of one function is just putting the complexity elsewhere, and potentially making the code more confusing (although the class structure is used a lot in this repo). I like the `export function abc(parameter, {pem, sdkId, apiUrl})` approach. If passing the same options gets tedious, you could create a wrapper function:
```
const wrapper = ({pem, sdkId, apiUrl}) => {
  return {
    abc: (parameter) => { ... },
  }
}
```

laurent-yoti:
@michaelyoti as much as using the `export function abc(parameter, {pem, sdkId, apiUrl})` is possible, it then requires more refactoring, and is potentially a breaking change for users who are directly using the method `abc`.
The choice of having class for services was also motivated by the alignment it provides with the DocScanService, which is already set up this way. Additionally, we could imagine creating a parent class `AbstractYotiService` that all services extends, so the "complexity" would be handled in one place only.
I am a bit dubious of creating a wrapper like because this is essentially achieving the same goal of the class and can be less obvious.

PR #328, opened 2022-09-01
SDK-2171 Deprecate old node support

abdalmajeed-yoti:
Updated:
package.json to drop support for versions older than 14.
tests.yaml to drop testing for versions older than 14 and test for node versions 14, 16 and 18 only.
.travis.yaml to drop testing for versions older than 14 and test for node versions 14, 16 and 18 only.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


laurent-yoti (approved):
LGTM
Acknowledging that the Node 18 tests are failing due to some dependencies that shall be updated in upcoming work.

PR #331, opened 2022-09-01
SDK-2176 Remove node rsa

liamd-yoti:
Replaced usages of `node-rsa` with Node's `crypto`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "bytebuffer": "^5.0.1",
         "husky": "^8.0.1",
         "node-forge": "^1.3.0",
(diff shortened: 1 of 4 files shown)


laurent-yoti (commented):
I was actually maybe expecting this one to be using `node-forge`... but we would anyway have to first update it to see.
It goes back to using the native `crypto` module or to rely on `node-forge` to safe-gard us from API changes...
Anyhow, what you did looks good.
Let's see once we have the updated `node-forge` in what would it look like with it, and then decide.

laurent-yoti (approved):
Looks good to me, seems there are some conflict, could you have a look please

PR #332, opened 2022-09-01
SDK-2177 Remove usages of Buffer from safe-buffer

liamd-yoti:
Replace `safe-buffer` usages with Node's built in `Buffer`


laurent-yoti (approved):
LGTM, I thought there would have been some usage to change, but actually `safe-buffer` was already having the same implementation/signature than the Node API!

PR #333, opened 2022-09-01
SDK-2174 Upgrade examples dependencies

abdalmajeed-yoti:


laurent-yoti:
Could you update the example to also mention 
```
"engines": {
    "node": ">=14"
},
```
They all do have different versions now.

laurent-yoti (commented):
LGTM overall.
We really would need to stop having all the examples redefining each dependencies, maybe looking into some npm workspace, so to avoid inconsistencies.

laurent-yoti (approved):
LGTM, please merge.

PR #336, opened 2022-09-07
SDK-834 Remove getUserId

abdalmajeed-yoti:


laurent-yoti (approved):
Looks good to me.
@dhivya-pattabhiraman could you have a look and confirm that the examples are not affected please.

dhivya-pattabhiraman (approved):
getUserId method is not used in the profile examples.

PR #340, opened 2022-09-08
SDK-892 Change DOB from string to date

liamd-yoti:
`getDateOfBirth` now returns `Date` instead of string.

### DIFF
--- a/tests/yoti_common/converters/attribute.converter.spec.js
+++ b/tests/yoti_common/converters/attribute.converter.spec.js
@@ -241,7 +241,13 @@ describe('AttributeConverter', () => {
     });
     it('should return Date values', () => {
       const converted = AttributeConverter.convertValueBasedOnContentType('1979-01-01', CONTENT_TYPE_DATE);
+      const expectedYear = 1979;
+      const expectedMonth = 0;
+      const expectedDay = 1;
       expect(converted).toBeInstanceOf(Date);
+      expect(converted.getFullYear()).toBe(expectedYear);
+      expect(converted.getMonth()).toBe(expectedMonth);
+      expect(converted.getDay()).toBe(expectedDay);
     });
   });
   describe('#convertValueBasedOnAttributeName', () => {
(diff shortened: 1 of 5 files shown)


abdalmajeed-yoti:
How about converting it to a `Date` object after it's converted in the `AttributeConverter` instead of converting it every time `getDateOfBirth` gets called? 
I think you can do that in `/yoti-node-sdk/src/yoti_common/converters/attribute.converter.js` line `51`.

liamd-yoti:
> How about converting it to a `Date` object after it's converted in the `AttributeConverter` instead of converting it every time `getDateOfBirth` gets called? I think you can do that in `/yoti-node-sdk/src/yoti_common/converters/attribute.converter.js` line `51`.

Okay think I was confused about this. It would indeed work. However when I was running the unit-tests for `profile.spec.js`, they are actually not a source of truth for the data types as we construct just the `Profile` class which does not perform any sort of conversion itself. The conversion is only done when we decrypt some profile. So yeah just updated the tests for the converter so that we can confirm that the data type is indeed changed when we receive some date field.

abdalmajeed-yoti (approved):
Looks good to me. 👍

tests/yoti_common/converters/attribute.converter.spec.js
```diff
@@ -239,6 +239,10 @@ describe('AttributeConverter', () => {
         expect(errMessage).toBe('Warning: value is NULL', `Content Type: ${contentType}`);
       });
     });
+    it('should return Date values', () => {
+      const converted = AttributeConverter.convertValueBasedOnContentType('1979-01-01', CONTENT_TYPE_DATE);
+      expect(converted).toBeInstanceOf(Date);
```
laurent-yoti, line 244: Probably good to check the date is matching ;-)
Please add a few lines to check the yeah, month...

abdalmajeed-yoti (approved):

laurent-yoti (approved):
Great, please merge.

PR #349, opened 2022-09-12
SDK-1271 Refactor profile and attributes handling

laurent-yoti:
Refactor profile/attributes:
- activity.details: removed deprecated methods
- profile: removed deprecated methods, simplified handling of age verification attribute
- attributes conversion, the value of an age verification is an instance of AgeVerification

Deleted deprecated Age class

Updated `examples/profile` to use supported method only and clearly list them.


src/data_type/age.verification.js
```diff
@@ -1,24 +1,30 @@
...
-    Validation.matchesPattern(attribute.getName(), /^[^:]+:(?!.*:)[0-9]+$/, 'attribute.name');
-    this.attribute = attribute;
+  static isAttributeNameMatchingAgeVerification(name) {
+    return name.startsWith(constants.ATTR_AGE_OVER)
+      || name.startsWith(constants.ATTR_AGE_UNDER);
+  }
+
+  constructor(name, value) {
+    Validation.isString(name, 'name');
+    Validation.oneOf(value, ['true', 'false'], 'value');
```
liamd-yoti, line 20: Can we not use `Validation.isBoolean`?

examples/profile/controllers/profile.controller.js
```diff
@@ -83,16 +171,21 @@ module.exports = (req, res) => {
...
 
   const promise = yotiClient.getActivityDetails(token);
   promise.then((activityDetails) => {
     const profile = activityDetails.getProfile();
+    const attributes = extractAttributesFromProfile(profile);
+    const displayableAttributes = attributes.map(toDisplayableAttribute)
+      .filter((attribute) => !!attribute);
+
+    // DEBUG
+    // console.log('>>>>\n', JSON.stringify(displayableAttributes, null, 2), '\n<<<< displayableAttributes');
```
liamd-yoti, line 184: Note: remove this?

src/yoti_common/index.js
```diff
@@ -61,6 +61,25 @@ function decryptEncryptedData(encryptedData, wrappedReceiptKey, pem) {
...
+ */
+function decryptProfileContent(profileContent, wrappedReceiptKey, pem) {
+  const profileNotEmpty = profileContent && Object.keys(profileContent).length > 0;
+
+  if (profileNotEmpty) {
+    const decryptedProfileContent = decryptEncryptedData(profileContent, wrappedReceiptKey, pem);
+    const { attributes: rawAttributes } = messages.decodeAttributeList(decryptedProfileContent);
+    return { attributes: AttributeListConverter.convertAttributeList(rawAttributes) };
+  }
+  console.log('Profile data is empty');
```
liamd-yoti, line 79: remove this log?

tests/profile_service/index.spec.js
```diff
@@ -33,20 +33,19 @@ describe('profileService', () => {
...
-
-            expect(profile).not.toBe(undefined);
             expect(receipt.getReceiptId()).toBe(receiptId);
             expect(receipt.getRememberMeId()).toBe(rememberMeId);
             expect(receipt.getParentRememberMeId()).toBe(parentRememberMeId);
             expect(receipt.getTimestamp()).toBeInstanceOf(Date);
             expect(receipt.getTimestamp().toUTCString()).toBe('Tue, 19 Jul 2016 08:55:38 GMT');
-            expect(profile.phoneNumber).toBe(phoneNumber);
-            expect(`data:image/jpeg;base64,${profile.selfie.toString('base64')}`).toBe(selfie);
-            expect(receipt.getBase64SelfieUri()).toBe(selfie);
```
liamd-yoti, line 47: Are we not testing this one anymore?

tests/profile_service/index.spec.js
```diff
@@ -66,16 +65,17 @@ describe('profileService', () => {
       it('should get an empty receipt from an empty profile share', (done) => {
         getReceipt('a-test-token', privateKeyFile, 'stub-app-id')
           .then((receipt) => {
```
liamd-yoti, line 67: Since we are refactoring here, can we change this to `async` instead of using `done()`?

tests/sample-data/profile-service/profile.json
```diff
@@ -1,32 +1,33 @@
...
   {
     "name": "phone_number",
     "value": "+447474747474",
     "sources": [],
     "verifiers": []
   },
-  "given_names":
+  {
+    "name": "date_of_birth",
+    "value": "2080-09-11T10:05:08.000Z",
```
liamd-yoti, line 21: Not sure that `value` here would be in that format? I was under the impression that it would be like `YYYY-MM-DD`

src/data_type/age.verification.js
laurent-yoti: The actual value of the attribute is a string... agreed this is not the best, but that is what we received for it in the content type.

examples/profile/controllers/profile.controller.js
laurent-yoti: Yeah, will be, examples are still to be finalised.

tests/profile_service/index.spec.js
laurent-yoti: The profile is exposing sugar methods to access each attribute, or you can get the list of attributes... but there is no longer any properties on the profile like `profile.phoneNumber` (mainly due to the support of multiple attribute sharing the same name).

Regarding `receipt.getBase64SelfieUri()`, that is deprecated, the selfie is an attribute, it never belonged to the receipt directly.

tests/profile_service/index.spec.js
laurent-yoti: Yes... but not now!
We ought to revisit all tests, probably also using `eslint/jest:recommended` which would flag this amongst other things.

tests/sample-data/profile-service/profile.json
laurent-yoti: Yeah, good shout!
Although, that `profile.json` fixture is a bit odd, it represents the decoded attributes list, where some attributes have been converted to non JSON type (aka class or type like Date), hence we need to revive them... and therefore, the value here being a date in ISO format is actually kinda valid, as it will be revived as a Date... but both are valid overall.

liamd-yoti: Yeah might just be worth keeping at as `YYYY-MM-DD` if that is what the API returns, just for consistency.

tests/profile_service/index.spec.js
liamd-yoti: Okay cool, thanks for the explanation.

src/data_type/age.verification.js
liamd-yoti: Okay interesting, seems fine then if that is what we get from the API. Although we could coax those values similarly to how we are doing with dates. Though given it is of string type, it would be slightly more complex to achieve that. Ie we would have to check if the string is either `"true" || "false"` which may be a little bit magic. Anyway, don't want to introduce too much complexity here, so I think its probably fine to keep these string variants for now.

abdalmajeed-yoti, line 20: Shouldn't the constructor be written first?

laurent-yoti: Yeah, the main thing to solve is how we consume and transform:
- from a proto attribute, we should convert to what this content type is saying.
- from the converted attribute to its type, we could then convert to more business logic related type, and indeed coerced such fake boolean.

To be reviewed one day.

tests/sample-data/profile-service/profile.json
laurent-yoti: Updated

src/data_type/age.verification.js
laurent-yoti: I don't think there is any rules around that, I could not find anything after 5min of googling, feel free to provide some documentation if you have.
I think this is only a matter of style, I am used to see them on top, as the first thing a class offers, before reading what is related to the instance behaviour (getter/setter/constructor/instance methods).

abdalmajeed-yoti: Yes, I couldn't find any doc or article about this.
But there are two other classes in the node SDK which write the constructor before static.

src/yoti_common/index.js
laurent-yoti: Yup, wiped out!

src/data_type/age.verification.js
laurent-yoti: Ok, so let's go with that.
We may follow stricter convention when we go Typescript.

PR #359, opened 2022-11-28
SDK-1975 Support for face capture resource

abdalmajeed-yoti:
Add support for:

- Creating a face capture resource
- Retrieving a face capture response
- Uploading a face capture image


src/idv_service/session/create/face_capture/upload.face.capture.image.payload.builder.js
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @class UploadFaceCaptureImagePayloadBuilder
+ */
+class UploadFaceCaptureImagePayloadBuilder {
+  /**
+   * Sets the content type for uploading a JPEG image
+   *
+   * @return this
+   */
+  forJpegImage() {
```
liamd-yoti, line 16: Do we really need these `for` methods?

Just wondering if that could become a problem for a consumer.

i.e they use an image like `cat.png` and accidentally use the `.forJpegImage()`. What would happen?

One alternative to these methods is to instead to use https://www.npmjs.com/package/file-type, to detect the file-type from the image buffer itself. That way you are kinda removing the need for these methods. Then it would just be a case of checking if the mimeType is a valid one (image/jpg or image/png).

abdalmajeed-yoti: I do agree with you and I wanted to implement an internal file type checker instead of these methods, but I think we need to follow the pattern of the existing same PRs for other languages.

package.json
```diff
@@ -29,6 +29,7 @@
     "coverageDirectory": "./coverage"
   },
   "dependencies": {
+    "form-data": "^4.0.0",
```
michaelyoti, line 32: It seems like we only use exact versions for packages here. (Maybe we should specify that in an .npmrc)

abdalmajeed-yoti: Yes, I think it would be good to specify that in an `.npmrc`.
But I'm not sure if we should add it in this PR or in a separate one.

src/idv_service/session/retrieve/face.capture.resource.response.js
```diff
@@ -0,0 +1,29 @@
...
+class FaceCaptureResourceResponse extends ResourceResponse {
+  constructor(resource) {
+    super(resource);
+
+    const { image, source } = resource;
+
+    Validation.isPlainObject(image, 'image');
+    this.image = new FaceCaptureImageResponse(image);
+
+    Validation.isPlainObject(source);
```
michaelyoti, line 17: Is this missing a second argument?

src/idv_service/session/create/face_capture/create.face.capture.resource.payload.builder.js
```diff
@@ -0,0 +1,28 @@
...
+ */
+class CreateFaceCaptureResourcePayloadBuilder {
+  /**
+   * Sets the id of the requirement that the resource will be used to satisfy.
+   *
+   * @param string requirementId
+   * @return CreateFaceCaptureResourcePayloadBuilder
+   */
+  withRequirementId(requirementId) {
+    Validation.isString(requirementId, 'requirement_id');
```
laurent-yoti, line 18: In every builder that have been added in this PR, the patter for the validation is passing as last parameter the name of the field... so the any error message could include the name of the field.
That is quite useful when validating an object and its internal fields, less so when actually validating some parameters of a method... but for the little benefit it provides, I think we would need to re-use the same name as the parameter (see camel case vs snake case). I have check in the rest of the project, and the rules seems to use the name of the parameter, ie `requirementId` here.

Overall, it would be better to have a nicer message, which let you know that it failed with a parameter {insert name of parameter} for a given method... but maybe we will simply just change the code to improve validation another time so we would not do that.

src/request/request.builder.js
```diff
@@ -176,12 +220,18 @@ class RequestBuilder {
     // Build endpoint and url.
     const endpointPath = `${this.endpoint}?${queryString}`;
 
+    const requestCanSendPayload = yotiCommon.requestCanSendPayload(this.method);
```
laurent-yoti, line 223: Just a bit weird to reuse the same name as the method for that const, isn't `canSendPayload` good enough?

src/idv_service/session/create/face_capture/create.face.capture.resource.payload.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @class CreateFaceCaptureResourcePayload
+ */
+class CreateFaceCaptureResourcePayload {
+  /**
+   * @param {string} requirementId
+   */
+  constructor(requirementId) {
+    Validation.isString(requirementId, 'requirementId');
+    this.requirement_id = requirementId;
```
laurent-yoti, line 15: Keep it camel case.
That is not related directly to the payload, is it ?

laurent-yoti: As in, there should be somewhere some conversion... will check more.

src/client/idv.client.js
```diff
@@ -93,6 +93,28 @@ class IDVClient {
     return this.idvService.getSupportedDocuments();
   }
 
+  /**
+   * @param {string} sessionId
+   * @param {CreateFaceCaptureResourcePayload} createFaceCaptureResourcePayload
+   *
+   * @returns {Promise<CreateFaceCaptureResourceResponse>}
+   */
+  createFaceCaptureResource(sessionId, createFaceCaptureResourcePayload) {
```
michaelyoti, line 102: Should we add a description for these new methods like the other ones have?

examples/idv/index.js
```diff
@@ -24,6 +24,7 @@ app.use(session({
 const router = express.Router();
 
 router.get('/', controllers.indexController);
+router.get('/fc', controllers.indexControllerWithFaceComparison);
```
liamd-yoti, line 27: Why `index`?

The other routes are not using that naming minus the actual index with is `/`

examples/idv/src/controllers/index.controller.with.face.comparison.js
```diff
@@ -0,0 +1,117 @@
...
+
+  return idvClient.createSession(sessionSpec);
+}
+
+async function addFaceCaptureResourceToSession(sessionId) {
+  const idvClient = new IDVClient(config.YOTI_CLIENT_SDK_ID, config.YOTI_PEM);
+
+  const sessionConfiguration = await idvClient.getSessionConfiguration(sessionId);
+
+  console.log('>>>>', sessionConfiguration, '<<<< sessionConfiguration');
```
liamd-yoti, line 68: Do we want to keep all these logs around?

src/client/idv.client.js
```diff
@@ -93,6 +93,28 @@ class IDVClient {
...
+
+  /**
+   * @param {string} sessionId
+   * @param {string} resourceId
+   * @param {UploadFaceCaptureImagePayload} uploadFaceCaptureImagePayload
+   *
+   * @returns {Promise<CreateFaceCaptureResourceResponse>}
+   */
+  uploadFaceCaptureImage(sessionId, resourceId, uploadFaceCaptureImagePayload) {
+    // eslint-disable-next-line max-len
```
liamd-yoti, line 114: minor:

Just wrap the arguments onto separate lines to avoid the `eslint-disable`

src/idv_service/idv.service.js
```diff
@@ -229,6 +233,66 @@ class IDVService {
     });
   }
 
+  /**
+   * @param {string} sessionId
+   * @param {CreateFaceCaptureResourcePayload} createFaceCaptureResourcePayload
+   *
+   * @returns {Promise} Resolves CreateFaceCaptureResourceResponse
```
liamd-yoti, line 240:
```suggestion
   * @returns {Promise<CreateFaceCaptureResourceResponse>}
```

liamd-yoti: Matter of style, elsewhere we had been using the angled brackets syntax to show the resolved type

michaelyoti:
Otherwise LGTM

tests/request/payload.spec.js
```diff
@@ -1,24 +1,86 @@
...
+
+    describe('#getContentType', () => {
+      it('should return the payload content type', () => {
+        const payloadJSON = payload.getContentType();
+        expect(payloadJSON).toBe('application/json');
+      });
+    });
+
+    describe('#getPayloadData', () => {
+      it('should return the payload data as a string', () => {
```
liamd-yoti, line 26: Just wondering here if you want to assert that what we got back is actually a string that reads how we expect. You have access here to the actual object you passed to the Payload (`const json = { a: '1' }`).

liamd-yoti (approved):
Just a few comments about style etc, but they are mostly minor things.

Otherwise LG 👍

examples/idv/src/controllers/index.controller.with.face.comparison.js
abdalmajeed-yoti: I think these were adding for testing only and can be removed, right @laurent-yoti ?

laurent-yoti: yeah, just temporary, let's remove them

examples/idv/index.js
laurent-yoti: You are right, this was only for checking quickly.
Before merging, we'll check with Dhivya how we want the examples to be.
The current example works with the index directly generating the session... which does not give room for configuring the session... so, yeah, maybe we should refactor so from the index you can configure the session, and therefore having only one route?

examples/idv/src/controllers/index.controller.with.face.comparison.js
```diff
@@ -0,0 +1,117 @@
...
+
+  const faceCaptureResourceRequirements = sessionConfiguration.getCapture().getFaceCaptureResourceRequirements();
+
+  if (faceCaptureResourceRequirements.length) {
+    const [firstRequirement] = faceCaptureResourceRequirements;
+    console.log('>>>>', firstRequirement.getId(), '<<<< firstRequirement.getId()');
```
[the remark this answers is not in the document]
laurent-yoti: Nope nope nope.

tests/idv_service/session/retrieve/allowed.source.response.spec.js
```diff
@@ -0,0 +1,22 @@
+const AllowedSourceResponse = require('../../../../src/idv_service/session/retrieve/allowed.source.response');
```
laurent-yoti, line 1: Hum...
We already have a class called `allowed.source.response.js` in `/src/idv_service/session/retrieve/allowed.source.response.js` , so why are we introducing a new one ?

tests/idv_service/session/retrieve/allowed.source.response.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('AllowedSourceResponse', () => {
+  let extendedAllowedSourceResponse;
+
+  beforeEach(() => {
+    class ExtendedAllowedSourceResponse extends AllowedSourceResponse {
+      constructor(type) {
+        super();
+        this.type = type;
+      }
```
laurent-yoti, line 11: No need to defined the constructor, that will be inherited, and that is exactly what we want to test here, ie, does the implementation in the abstract doing the right thing, so it would be counter productive to have an implementation in the sub class.

tests/request/payload.spec.js
```diff
@@ -1,24 +1,86 @@
...
+        const payloadData = payload.getPayloadData();
+        expect(payloadData).toEqual(JSON.stringify(json));
+      });
+    });
+
+    describe('#getBase64Payload', () => {
+      it('should return the payload data as base64 string', () => {
+        const base64Payload = payload.getBase64Payload();
+        expect(base64Payload).toBe(expectedBase64JsonPayload);
```
laurent-yoti, line 35: Seems weird to see `expectedBase64JsonPayload` defined above where the actual `json` is defined.
Surely they should be in relation?

tests/request/payload.spec.js
```diff
@@ -1,24 +1,86 @@
...
+      });
     });
   });
 
-  describe('#getBase64Payload', () => {
-    it('should return the base64Payload string', () => {
-      const base64Payload = amlPayload.getBase64Payload();
-      expect(base64Payload).toBe(expectedBase64Payload);
+  describe('when content type is data form', () => {
+    const dataFormFields = {
```
laurent-yoti, line 41: I struggled to understand what `dataFormFields` was.
So, that is the payload data parameter... not sure what would be a better name...  `payloadData` maybe ?

tests/request/payload.spec.js
```diff
@@ -1,24 +1,86 @@
...
+
+        const formData = new FormData();
+        formData.setBoundary(payload.getRawData().getBoundary());
+        const { name, value } = dataFormFields.getFormDataFields()[0];
+        formData.append(name, value);
+
+        const expectedBase64 = formData.getBuffer().toString('base64');
+
+        expect(base64Payload).toBe(expectedBase64);
+      });
```
laurent-yoti, line 83: Not sure about these tests.
These are not easy, but here, you are way a bit repeating the implementation...

We could mock `FormData` so we can control what it returns when calling its constructor, so we can make sure that we control what `.getBuffer()` returns, we can also assert that a new FormData is created, and asserting that its `.append()` method is called with the right data ?

tests/idv_service/session/retrieve/allowed.source.response.spec.js
```diff
@@ -0,0 +1,22 @@
+const AllowedSourceResponse = require('../../../../src/idv_service/session/retrieve/allowed.source.response');
+
+describe('AllowedSourceResponse', () => {
+  let extendedAllowedSourceResponse;
+
+  beforeEach(() => {
+    class ExtendedAllowedSourceResponse extends AllowedSourceResponse {
```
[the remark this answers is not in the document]
abdalmajeed-yoti: Has been removed and left the other used one.

tests/idv_service/session/retrieve/allowed.source.response.spec.js
```diff
@@ -0,0 +1,22 @@
...
+      constructor(type) {
+        super();
+        this.type = type;
+      }
+    }
+
+    extendedAllowedSourceResponse = new ExtendedAllowedSourceResponse('one');
+  });
+
+  describe('#getType', () => {
```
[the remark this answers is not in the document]
abdalmajeed-yoti: Has been removed and left the other used one. and that case is tested in the other one.

tests/request/payload.spec.js
abdalmajeed-yoti: Yes, but I was thinking of making the test like the ones for json, to have a constant value and tests if the output is the same.
The only issue I had was that the boundary is for the form data is dynamic, so with the new changes I made it static.

PR #362, opened 2022-12-07
SDK-1985 Fetching session config support

abdalmajeed-yoti:
Support for fetching session configuration.


src/client/idv.client.js
```diff
@@ -92,6 +92,18 @@ class IDVClient {
...
     return this.idvService.getSupportedDocuments();
   }
+
+  /**
+   * Fetches the configuration for the given sessionID.
+   *
+   * @param {string} sessionId
+   * @return {SessionConfigurationResponse}
+   *
+   * @returns {Promise}
```
laurent-yoti, line 102: I used a bit the `examples/idv` to try it out, works fine!
But I realised that we are not really doing a good job with the JSDoc when it comes to promise.

Could you please update the `idv.client.js` and `idv.service.js` JSDoc comments so:
- there is only 1 `@return`
- use `return` or `returns` (which ever is the one the most used in the project)
- replace any `@returns {Promise}` by `@returns {Promise<TheClassItResolvesWith>}`

I had a look about the reject/error, but there is nothing available in JSDoc it seems... so, we'll get these when on Typescript later.

src/idv_service/session/retrieve/configuration/capture/required.resource.response.js
```diff
@@ -0,0 +1,97 @@
...
+   * Returns if the Relying Business is allowed to upload resources
+   * to satisfy the requirement.
+   *
+   * return the end user is allowed to upload resources
+   *
+   * @return {boolean}
+   */
+  isRelyingBusinessAllowed() {
+    if (this.allowedSources) {
+      return this.allowedSources.findIndex(
```
laurent-yoti, line 88: Could be replace with `.some`

tests/idv_service/session/retrieve/configuration/capture/document/objective.response.spec.js
```diff
@@ -0,0 +1,17 @@
+const ObjectiveResponse = require('../../../../../../../src/idv_service/session/retrieve/configuration/capture/document/objective.response');
+
+describe('ObjectiveResponse', () => {
+  let objectiveResponse;
+
+  beforeEach(() => {
+    objectiveResponse = new ObjectiveResponse({
+      type: '',
```
laurent-yoti, line 8: why keeping it as empty string ?

tests/idv_service/session/retrieve/configuration/session.configuration.response.spec.js
```diff
@@ -0,0 +1,44 @@
...
+
+describe('SessionConfigurationResponse', () => {
+  let sessionConfigurationResponse;
+
+  beforeEach(() => {
+    sessionConfigurationResponse = new SessionConfigurationResponse({
+      client_session_token_ttl: 123,
+      session_id: 'some_id',
+      requested_checks: [
+        '',
```
laurent-yoti, line 12: That seems bizarre... should we have a least 1 check, so then we can verify the length and the type of check it was converted to ?

tests/idv_service/session/retrieve/configuration/capture/document/required.supplementary.document.resource.response.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  let requiredSupplementaryDocumentResourceResponse;
+
+  beforeEach(() => {
+    // eslint-disable-next-line max-len
+    requiredSupplementaryDocumentResourceResponse = new RequiredSupplementaryDocumentResourceResponse({
+      type: 'SUPPLEMENTARY_DOCUMENT',
+      id: '',
+      state: '',
+      document_types: [
+        '',
```
laurent-yoti, line 13: Same comment about empty values...

tests/idv_service/session/retrieve/configuration/capture/document/required.supplementary.document.resource.response.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      ],
+      country_codes: [
+        'GBR',
+      ],
+      objective: {
+        type: '',
```
laurent-yoti, line 19: Same comment about empty values...

tests/idv_service/session/retrieve/configuration/session.configuration.response.spec.js
abdalmajeed-yoti: There is no conversion for the requested checks.

src/idv_service/session/retrieve/configuration/capture/source/allowed.source.response.js
```diff
@@ -1,4 +1,11 @@
+const Validation = require('../../../../../../yoti_common/validation');
+
 class AllowedSourceResponse {
+  constructor(type) {
+    Validation.isString(type, 'type');
+    this.type = type;
+  }
```
laurent-yoti, line 7: So, that is fine, but this is not quite an Abstract class... which is not existing in vanilla js, but can be simulated... aka, the constructor of the Abstract throws an error if the class is instantiated, and therefore only allow its subclasses to be instantiated....
I am not sure of the value of having such Abstract behaviour, but this has been implemented in a few places in this code base.

tests/idv_service/session/retrieve/configuration/session.configuration.response.spec.js
laurent-yoti: ok

dhivya-pattabhiraman (approved):

liamd-yoti (approved):
👍

PR #363, opened 2022-12-12
SDK-1985 fetching session config support example

dhivya-pattabhiraman:
Updated the IDV example to retrieve and display the session configuration in a new tab.

### DIFF
--- a/examples/idv/views/pages/index.ejs
+++ b/examples/idv/views/pages/index.ejs
@@ -1,3 +1,7 @@
 <%- include('layout/header'); -%>
+<div style="font-size: 18px; margin-left: 300px">
+    <span>Session ID:</span> <span><%= sessionId %></span>
+</div>
+<a style="font-size: 18px; margin-left: 300px" href="/config" target="_blank">View session configuration</a>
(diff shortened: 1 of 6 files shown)


examples/idv/views/pages/index.ejs
```diff
@@ -1,3 +1,7 @@
 <%- include('layout/header'); -%>
+<div style="font-size: 18px; margin-left: 300px">
+    <span>Session ID:</span> <span><%= sessionId %></span>
+</div>
+<a style="font-size: 18px; margin-left: 300px" href="/config" target="_blank">View session configuration</a>
```
laurent-yoti, line 5: Is that on purpose that the href is not including the `sessionID` query params ?

abdalmajeed-yoti (approved):

examples/idv/views/pages/index.ejs
dhivya-pattabhiraman: Yeah, the sessionID was not needed in the href.

dhivya-pattabhiraman: Actually added the query param.

laurent-yoti (approved):
Looks great, please merge.

PR #370, opened 2023-02-09
SDK-2237 share v2 create session

abdalmajeed-yoti:
Added `createSession` for creating a share session.


src/digital_identity_service/index.js
```diff
@@ -9,6 +13,45 @@ class DigitalIdentityService {
...
     this.apiUrl = apiUrl;
   }
+
+  /**
+   *
+   * @param shareSessionConfig
+   * @returns {Promise<ShareSessionResult>}
+   */
+  createSession(shareSessionConfig) {
+    const jsonShareSession = JSON.parse(JSON.stringify(shareSessionConfig));
```
liamd-yoti, line 23: I think we also need to add some builders for the session config. I had done that in the draft PR, `ShareSessionBuilder` and `ShareSessionNotificationBuilder`. please check those.

src/digital_identity_service/index.js
```diff
@@ -9,6 +13,45 @@ class DigitalIdentityService {
...
+
+    const payload = new Payload(jsonShareSession);
+
+    const requestBuilder = new RequestBuilder()
+      .withBaseUrl(this.apiUrl)
+      .withHeader('X-Yoti-Auth-Id', this.sdkId)
+      .withPemString(this.pem)
+      .withEndpoint('/v2/sessions')
```
liamd-yoti, line 31: `/v2` is not needed anymore since the apirUrl is encapsulating that. You can just have `/sessions` that will be enough.

src/digital_identity_service/index.js
```diff
@@ -9,6 +13,45 @@ class DigitalIdentityService {
...
+      .withQueryParam('appId', this.sdkId)
+      .withMethod('POST')
+      .withPayload(payload);
+
+    const request = requestBuilder.build();
+
+    return new Promise((resolve, reject) => {
```
liamd-yoti, line 38: Maybe its personal opinion but I don't see a good reason for not having these class methods as `async` and handling errors in `try` `catch`. Maybe there is a reason for this though.

abdalmajeed-yoti: I do agree with you, so yes maybe in this new service we should follow the pattern of `async` with `try` `catch`.
What do you think @laurent-yoti , @michaelyoti , @liamd-yoti ?

src/digital_identity_service/share.session.result.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+const Validation = require('../yoti_common/validation');
+
+/**
+ * The share session result
+ *
+ * @class ShareSessionResult
+ */
+module.exports = class ShareSessionResult {
```
laurent-yoti, line 10: What is the intent with this class regarding the result of a fetch session ?
Should there be different ? Probably yes to make things simpler, but if so, what name would we go with ?

src/digital_identity_service/share.session.result.js
```diff
@@ -0,0 +1,53 @@
...
+  constructor(response) {
+    Validation.isString(response.id, 'Session ID');
+    this.id = response.id;
+
+    Validation.isString(response.status, 'Status');
+    this.status = response.status;
+
+    Validation.isString(response.expiry, 'Expiry');
+    const expiryDate = new Date(response.expiry);
+    if (expiryDate.toString() === 'Invalid Date') throw TypeError('Expiry must be a date like string');
```
laurent-yoti, line 23: Given we are going to have other payload with date, let's move that to the utils.

src/digital_identity_service/index.js
laurent-yoti: Consistency wise, it is easier to keep one style... and then change them all once we move to TS.
Yet, we can already considering "educating" users with `async/await` now.

laurent-yoti:
Could you please add the tests.

We are probably going to hit a issue with the duplication strategy regarding the code quality inspection... let's see if there is a way to flag that as being acceptable for a sub section of the code base.

dhivya-pattabhiraman:
Hello @liamd-yoti and @laurent-yoti . I went thro' this PR and I have few questions. We have copied over the Dynamic policy files and the Extensions files to the new Digital Identity Service but we have not exposed the methods to create a dynamic policy or extensions in the Digital Identity Service. So when the user creates an instance of say dynamicPolicyBuilder they will be creating an instance of the class that resides within dynamic_sharing_service and not digital_identity_service and this will cause issues when Validating. You can check the main index.js file. The dynamic_sharing_service is importing and exporting a lot of builders. We would need a similar setup for the Digital Identity Service as well.

abdalmajeed-yoti:
@dhivya-pattabhiraman , we were discussing that and agreed to export the builders in an additional namespace.
So I'll export the builders in a new namespace.

src/digital_identity_service/share.session.result.js
dhivya-pattabhiraman: I also agree with different names like ShareSessionCreateResult and ShareSessionFetchResult OR CreateShareSessionResult and FetchShareSessionResult

src/yoti_common/validation.js
```diff
@@ -30,6 +30,25 @@ module.exports = class Validation {
...
+   */
+  static isStringDate(value, name, optional = false) {
+    if ((typeof value) === 'undefined' && optional) {
+      return;
+    }
+
+    const date = new Date(value);
+
+    if (date.toString() === 'Invalid Date') {
+      throw TypeError('Expiry must be a date like string');
```
laurent-yoti, line 48: Should not read `'Expiry...'` but the `name`.

tests/digital_identity_service/extension/extension.builder.spec.js
```diff
@@ -0,0 +1,21 @@
+const { ExtensionBuilder } = require('../../..');
+const Extension = require('../../../src/dynamic_sharing_service/extension/extension');
```
laurent-yoti, line 2: There a few files in the test referring to the `dynamic_sharing_service` version.
Also, I think some of these extensions are not created (copied) in the digital identity version.

src/digital_identity_service/index.js
```diff
@@ -1,16 +1,103 @@
...
+      .withPemString(this.pem)
+      .withEndpoint('/v2/sessions')
+      .withQueryParam('appId', this.sdkId)
+      .withMethod('POST')
+      .withPayload(payload);
+
+    const request = requestBuilder.build();
+
+    try {
+      const response = await request.execute().catch((error) => {
```
michaelyoti, line 57: Is it correct to still have the `catch` here? Isn't the `try`/`catch` going to handle this now, since there is an `await`?

abdalmajeed-yoti: This chained `catch` is to handle the `execute` error, but the next `catch` is for handling the error of creating an instance of `ShareSessionCreateResult`. 

So I had to chain `catch` with `execute` or have nested `try` `catch`.

What do you think, should we do it this way or have a nested `try` `catch`?

src/client/digital.identity.client.js
```diff
@@ -17,6 +17,15 @@ class DigitalIdentityClient {
 
     this.digitalIdentityService = new DigitalIdentityService(sdkId, pem, options);
   }
+
+  /**
+   *
+   * @param shareSessionConfig
+   * @returns {Promise<ShareSessionResult>}
```
laurent-yoti, line 24: Should be `ShareSessionCreateResult`

src/digital_identity_service/index.js
```diff
@@ -1,16 +1,103 @@
...
     this.pem = pem;
     this.apiUrl = apiUrl;
   }
+
+  /**
+   *
+   * @param shareSessionConfig
+   * @returns {Promise<ShareSessionCreateResult>}
+   */
+  async createSession(shareSessionConfig) {
```
laurent-yoti, line 40: Let's call it `createShareSession`.

src/digital_identity_service/index.js
michaelyoti: Well, if the chained `catch` is handling the error from `execute`, why is it in the `try` block?

Then, I'm not sure how the chained `catch` works synchronously. What happens if `execute` errors, and goes to into the chained `catch`? Will it throw that error in the main thread?

src/digital_identity_service/index.js
```diff
@@ -1,16 +1,103 @@
...
     Validation.notNullOrEmpty(pem, 'pem');
 
     this.sdkId = sdkId;
     this.pem = pem;
     this.apiUrl = apiUrl;
   }
+
+  /**
+   *
+   * @param shareSessionConfig
```
laurent-yoti, line 37: This `shareSessionConfig`, it seems that this is an instance of `share.session`, which is itself build using `share.session.builder`.
The name is a bit problematic, as it looks like that this is a full fledged session, but really, this is only the configuration to create one.
In IDV, it is referred as `specification`, we could follow the same or use `configuration` ?

src/digital_identity_service/index.js
laurent-yoti: I think Majeed tried to respect the previous implementation, which is handling the call first, then the parsing of the data.
Yet, it is not quite equivalent to what existed (comparing with DynamicScenario service).

The .catch() will print the error like before and then throw the error, yet, this one is now caught by the try/catch, which will print the error again (yet with a different message), and finally throw it.

If we just want the same, we can simply use only try/catch, without nesting though.

```js
try {
  // do the call and assign result to variable
} catch(e) {
  // print error with associated message
 throw e
}

try {
  // convert result variable to final result class
} catch(e) {
  // print error with associated message
 throw e
}
```

tests/digital_identity_service/policy/dynamic.policy.builder.spec.js
```diff
@@ -0,0 +1,456 @@
+const {
```
laurent-yoti, line 1: Let's eliminate `Dynamic` from the file content and file name.

examples/digital-identity/README.md
```diff
@@ -0,0 +1,9 @@
+# Share session Example
+
+1. From the [Yoti Hub](https://hub.yoti.com) set the application domain of your app to `localhost:9443`
+1. Set the scenario callback URL to `/profile`
+1. Rename the [.env.example](.env.example) file to `.env` and fill in the required configuration values
+1. Install the dependencies with `npm install`
+1. Start the server `node index.js`
```
abdalmajeed-yoti, line 7: I think we should update this read me.

examples/digital-identity/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "1.0.0",
+  "description": "A Yoti digital identity client example",
+  "private": true,
+  "author": "Yoti LTD <tech@yoti.com> (https://www.yoti.com/developers)",
+  "license": "MIT",
+  "engines": {
+    "node": ">=14"
+  },
+  "scripts": {
+    "start": "nodemon index.js",
```
laurent-yoti, line 12: Just use node, or add `nodemon` as dev dependency.

examples/digital-identity/README.md
laurent-yoti: This indeed must be updated... we can leave it to the end, or do it now.
what do you prefer @dhivya-pattabhiraman ?

dhivya-pattabhiraman: I thought of updating it at the very end when we have a full-fledged example for share v2 normal session and identity profile check session.

liamd-yoti (approved):

dhivya-pattabhiraman (approved):

laurent-yoti (approved):

michaelyoti (approved):

PR #373, opened 2023-02-13
SDK-2243 share v2 retrieve session

abdalmajeed-yoti:
Added v2 `fetchSession`.

### DIFF
--- a/examples/digital-identity/controllers/share.session.controller.js
+++ b/examples/digital-identity/controllers/share.session.controller.js
@@ -1,3 +1,4 @@
+const { Router } = require('express');
 const {
   DigitalIdentityBuilders: {
     PolicyBuilder,
@@ -14,7 +15,11 @@ const digitalIdentityClient = new DigitalIdentityClient(
   config.PEM_KEY,
 );
 
-module.exports = (req, res) => {
+const router = Router();
+
+const jsonToFormattedString = (json) => JSON.stringify(json, undefined, 2);
+
+router.get('/createSession', (req, res) => {
   const locationExtension = new LocationConstraintExtensionBuilder()
     .withLatitude(51.5074)
     .withLongitude(-0.1278)
@@ -49,7 +54,7 @@ module.exports = (req, res) => {
   digitalIdentityClient
     .createShareSession(shareSessionConfig)
     .then((shareSessionCreateResult) => {
-      res.json({ ShareSessionResult: shareSessionCreateResult });
+      res.end(jsonToFormattedString({ ShareSessionResult: shareSessionCreateResult }));
     })
     .catch((error) => {
       const {
@@ -59,6 +64,16 @@ module.exports = (req, res) => {
         status, code, reason, message,
       };
       console.error(errorData);
-      res.json(errorData);
+      res.end(jsonToFormattedString(errorData));
     });
-};
+});
+
+router.get('/fetchSession/:sessionId', (req, res) => {
+  const { sessionId } = req.params;
+
+  digitalIdentityClient.fetchShareSession(sessionId).then((shareSessionFetchResult) => {
+    res.end(jsonToFormattedString(shareSessionFetchResult));
+  });
+});
+
+module.exports = router;
--- a/examples/digital-identity/index.js
+++ b/examples/digital-identity/index.js
@@ -18,7 +18,7 @@ app.use('/static', express.static('static'));
 
 const router = express.Router();
 
-router.get('/createSession', shareSessionController);
+router.use('/', shareSessionController);
 
 app.use('/', router);
 
--- a/src/digital_identity_service/index.js
+++ b/src/digital_identity_service/index.js
@@ -6,6 +6,7 @@ const Validation = require('../yoti_common/validation');
 const { RequestBuilder } = require('../request/request.builder');
 const { Payload } = require('../request/payload');
 const ShareSessionCreateResult = require('./share.session.create.result');
+const ShareSessionFetchResult = require('./share.session.fetch.result');
 const ShareQrCodeCreateResult = require('./share.qr.code.create.result');
 
 const WantedAnchorBuilder = require('./policy/wanted.anchor.builder');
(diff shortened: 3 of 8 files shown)


src/digital_identity_service/index.js
```diff
@@ -73,6 +74,40 @@ class DigitalIdentityService {
...
+      });
+
+      const parsedResponse = response.getParsedResponse();
+
+      return new ShareSessionFetchResult(parsedResponse);
+    } catch (err) {
+      console.log(`Error getting response data: ${err}`);
+
+      throw err;
+    }
```
laurent-yoti, line 109: Could you apply the same pattern as the creation, aka split `try/catch`...

examples/digital-identity/index.js
```diff
@@ -18,7 +18,7 @@ app.use('/static', express.static('static'));
 
 const router = express.Router();
 
-router.get('/createSession', shareSessionController);
+router.use('/', shareSessionController);
```
laurent-yoti, line 21: What is the intention with that ?
Is that just for session, and then we'd have another one for QR ?

examples/digital-identity/controllers/share.session.controller.js
```diff
@@ -59,6 +64,16 @@ module.exports = (req, res) => {
         status, code, reason, message,
       };
       console.error(errorData);
-      res.json(errorData);
+      res.end(jsonToFormattedString(errorData));
     });
-};
+});
+
+router.get('/fetchSession/:sessionId', (req, res) => {
```
laurent-yoti, line 71: Not really keen on that `fecthSession`...

Maybe we should have a chat about what we want to achieve for the example @dhivya-pattabhiraman ?

src/digital_identity_service/share.qr.code.create.result.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
```
laurent-yoti, line 1: What is that file about ??? This is related to the other PR ? Am I not meant to look at the retrieval of a session ?

src/digital_identity_service/share.session.fetch.result.js
```diff
@@ -0,0 +1,106 @@
...
+    this.updated = new Date(response.updated);
+
+    if (response.expiry) {
+      Validation.isStringDate(response.expiry, 'Expiry');
+      this.expiry = new Date(response.expiry);
+    }
+
+    if (response.qrCode) {
+      Validation.isPlainObject(response.qrCode, 'QrCode');
+      this.qrCode = new ShareQrCodeCreateResult(response.qrCode);
```
laurent-yoti, line 35: No!
We are only expecting an `id` to be in... and that would anyway not correspond to a creation....

src/digital_identity_service/share.session.fetch.result.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    if (response.receipt) {
```
laurent-yoti, line 38: Same as with the QR, if we have the receipt object, it shall only contain an id.
Let's validate the existence of `receipt.id`.

examples/digital-identity/index.js
abdalmajeed-yoti: Instead of creating routes inside `index.js` I thought of just making it use controllers.
> Is that just for session, and then we'd have another one for QR ?

Yes.

examples/digital-identity/controllers/share.session.controller.js
dhivya-pattabhiraman: So, what i had in mind was that when we have web share ready, we will create an example like we have for the Profile example that uses web share for integration (which will under the hood take care of QR code creation and refresh). And we only need to show Session creation and receipt retrieval. Fetching the session will be needed for getting the receipt id. What do you guys think?

laurent-yoti: Hum, I am not sure we'll have web-share ready there by then, also, using web-share would remove the need to use a few of the endpoint...
So, let's do something dry first...
Then, in the same example, once we have web-share, we can put in place another route that would illustrate how to use both of them.

dhivya-pattabhiraman (approved):

examples/digital-identity/controllers/share.session.controller.js
laurent-yoti: We could, or we could expose endpoints as dissociated...

src/digital_identity_service/share.session.fetch.result.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  /**
+   * The session receipt
+   *
+   * @returns {{id: string}} The session receipt
+   */
+  getReceipt() {
+    return this.receipt;
+  }
```
laurent-yoti, line 114: That is correct, yet, I am wondering whether we should simply store/expose their id directly.

So, we could have something like
```js
this.scannedQrCodeId = ''
this.receiptId = ''

// with something like
hasReceipt()
getReceiptId()

wasScanned()
getScannedQrCodeId()
```

What do you think ? @liamd-yoti @abdalmajeed-yoti @dhivya-pattabhiraman ?

abdalmajeed-yoti: Looks good

abdalmajeed-yoti: Do other SDKs need to follow this approach? or is it okay to apply it for the node SDK only?
Also I think we can just have these two methods only
```js
getReceiptId()
getScannedQrCodeId()
```
What do you think?

laurent-yoti: We would want the other SDK to be aligned too.
You are suggesting to remove the 2 extra methods, please provide some rational... like saying that using the `get` methods like is sufficient and is equivalent (minus the type conversion).

abdalmajeed-yoti: I think because these two methods are enough and the client can just check if the value is `null`.
```js
getReceiptId()
getScannedQrCodeId()
```

laurent-yoti: Ok, fine by me, let's do that then.

laurent-yoti (approved):
LGTM

PR #374, opened 2023-02-13
SDK-2261 Retrieve receipt

liamd-yoti:
Introduced method for getting and decrypting receipt

```js
const client = new DigitalIdentityClient(sdkId, pem);
const receipt = await client.fetchReceipt(receiptId);

receipt.getRememberMeId();
receipt.getProfile().getFullName();
...
```


src/digital_identity_service/index.js
```diff
@@ -201,6 +212,134 @@ class DigitalIdentityService {
...
+    try {
+      const response = await request.execute();
+      const parsedResponse = response.getParsedResponse();
+
+      return new ReceiptResponse(parsedResponse);
+    } catch (e) {
+      console.log(`fetching receipt error ${e}`);
+      throw e;
+    }
+  }
```
michaelyoti, line 240: This doesn't follow the pattern of the other methods, which have two `try` blocks. Otherwise, I don't see a problem with this approach. I'm actually not sure if the `try`s are necessary at all, since they just re-throw the error, and log something which I doubt will be useful to the user

src/client/digital.identity.client.js
```diff
@@ -18,6 +18,15 @@ class DigitalIdentityClient {
     this.digitalIdentityService = new DigitalIdentityService(sdkId, pem, options);
   }
 
+  /**
+   *
+   * @param {string} receiptId
+   * @returns {Promise<Receipt>}
+   */
+  async getReceipt(receiptId) {
```
michaelyoti, line 26: It seems inconsistent to be using `get` here while the other client methods use `fetch`. I know that we make a distinction between `getReceipt` and `fetchReceipt` at the service level but that doesn't seem like a clear way of differentiating those two methods. Maybe they should be called `fetchReceipt` and `fetchEncryptedReceipt` in `DigitalIdentityService`?

liamd-yoti: I agree on the `get < fetch` side of things. `getReceipt` should probably become `fetchReceipt`.

Regarding the other methods that I introduced here. I'm not even sure if we want to expose those on the client. If they are only used by this `getReceipt`. Do we think it's useful for integrators?

src/digital_identity_service/index.js
liamd-yoti: I agree. I think the approach from the other PRs has been to use the `DigitalIdentityServiceError` and throw that. It also contains some logic about networking errors. So should probably do the same here.

src/client/digital.identity.client.js
michaelyoti: Oh, I wasn't suggesting we expose the other methods on the client. I was just suggesting we rename the methods in `DigitalIdentityService` to be clearer and more consistent with this renaming in the client. So maybe,
* `DigitalIdentityService.getReceipt` -> `DigitalIdentityService.fetchAndDecryptReceipt` and
* `DigitalIdentityService.fetchReceiptById` -> `DigitalIdentityService.fetchEncryptedReceipt`

liamd-yoti: Ah ok.

Yes that sounds like a good idea. Then the exposed method via the client would be just `fetchReceipt`, since the details are not super important to an integrator.

dhivya-pattabhiraman:
@liamd-yoti @laurent-yoti I tested the fetch receipt PR locally and saw that in the response we have something like this,

{
  "receipt": {
    "id": "CtbL20ROQhgq925SdRKnC7aSl4K4TfqXP46cezuGdjBODcpTK7zmKtt1wQxrEByF",
    "sessionId": "ss.v2.ChZ3OEx3QUZ0OFRLS2tid3pzVWZIc0R3EgdsZDUuZ2Jy",
    "timestamp": "2023-02-22T13:27:05.000Z",
    "error": "B-UNEXPECTED-LOCATION"
  },
  "userProfile": {
    "attributes": [],
    "attributesMap": {},
    "ageVerifications": {}
  }
}

where the properties, id, sessionId, timestamp, error are wrapped inside receipt property. Can we avoid that? I feel we can have those properties in the root level and the userProfile property containing the attributes and related stuff. just my thought

michaelyoti (approved):

src/digital_identity_service/receipts/receipt.item.key.response.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+const Validation = require('../../yoti_common/validation');
+
```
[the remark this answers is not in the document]
liamd-yoti: Yeah in regards to that, I am not even sure we actually need these classes.

The methods that they correspond to are not exposed as part of the client. It seems that internally we could just use the JSON response from the method and not worry about "casting" them to some class. The only thing that class is really adding in terms of functionality is the validation aspect. Wondering if that is even useful?

In regards to your comment, yeah can add the JSdoc description and type. Although, it may not be super useful in the end given the class is not exposed.

src/digital_identity_service/receipts/receipt.response.js
```diff
@@ -0,0 +1,144 @@
...
+    Validation.isString(response.id, 'Receipt ID');
+    this.id = response.id;
+
+    Validation.isString(response.sessionId, 'Session ID');
+    this.sessionId = response.sessionId;
+
+    Validation.isString(response.timestamp, 'timestamp');
+    const timestamp = new Date(response.timestamp);
+    if (timestamp.toString() === 'Invalid Timestamp') throw TypeError('timestamp must be a date like string');
+    this.timestamp = timestamp;
```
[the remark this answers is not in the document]
liamd-yoti: Cool, will change it 👍

PR #392, opened 2023-06-29
SDK-2310 Add biometric consent flow

abdalmajeed-yoti:
Added `biometricConsentFlow` to the SDK config by using `withEarlyBiometricConsentFlow` or `withJustInTimeBiometricConsentFlow` methods.

### DIFF
--- a/src/idv_service/session/create/sdk.config.builder.js
+++ b/src/idv_service/session/create/sdk.config.builder.js
@@ -152,6 +152,37 @@ class SdkConfigBuilder {
   }
 
   /**
+   * Sets the biometric consent flow to "EARLY"
+   *
+   * @returns {this}
+   */
+  withEarlyBiometricConsentFlow() {
+    return this.withBiometricConsentFlow(IDVConstants.EARLY);
+  }
+
+  /**
+   * Sets the biometric consent flow to "JUST_IN_TIME"
+   *
+   * @returns {this}
+   */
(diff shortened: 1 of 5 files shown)


src/idv_service/session/create/sdk.config.builder.js
```diff
@@ -151,6 +151,37 @@ class SdkConfigBuilder {
...
+
+  /**
+   * Sets the biometric consent flow
+   *
+   * @param {string} biometricConsentFlow the biometric consent flow
+   *
+   * @returns {this}
+   */
+  withBiometricConsentFlow(biometricConsentFlow) {
+    Validation.isString(biometricConsentFlow, 'biometricConsentFlow');
```
laurent-yoti, line 180: Do we have any `oneOf` in the `Validation` ?
Feel that we could allow any value right now, and that method is by nature public.

MrBurtyyy: If you're implying validation on the value passed to this method by the Relying Business, this should be left to the backend.  It also allows older SDKs to provide new values that the backend may support, even if there isn't a "sugar" method for it.

laurent-yoti: Yup, thanks, as discussed, we shall not try to validate the value itself then.

laurent-yoti (approved):
Happy with the examples, let's merge.

PR #393, opened 2023-06-29
SDK-2338 enabling expanded document fields

abdalmajeed-yoti:
Added `withCreateExpandedDocumentFields` method to set a boolean value for `createExpandedDocumentFields` for the `RequestedTextExtractionTask`.

PR #397, opened 2023-06-30
SDK-2337 Support expanded document fields media

abdalmajeed-yoti:
Added support for expanded_document_fields media.


laurent-yoti (approved):
LGTM.
Let's add an example in, which also then illustrates #393

abdalmajeed-yoti:
Thank you @dhivya-pattabhiraman , it looks good.

PR #405, opened 2023-08-01
SDK-1975 support for face capture resource proposal

abdalmajeed-yoti:
Refactor proposal.


laurent-yoti (approved):
Nice one, let's use that.

PR #446, opened 2024-02-20
SDK-2427 update issue template

nikhilPank:


laurent-yoti:
LGTM!
Shall we target `master` branch directly, so it will be there from now on (we can rebase `development` just after)?

nikhilPank:
@laurent-yoti can do

PR #510, opened 2025-03-03
SDK-2359 Digital Identity - support wanted attributes extra options

laurent-yoti:
Added in `digital_identity_service` the wanted attribute extra options (`alternative name`, `optional`)

### DIFF
--- a/tests/digital_identity_service/policy/wanted.attribute.builder.spec.js
+++ b/tests/digital_identity_service/policy/wanted.attribute.builder.spec.js
@@ -131,4 +131,48 @@ describe('WantedAttributeBuilder', () => {
     expect(wantedAttribute).toBeInstanceOf(WantedAttribute);
     expect(JSON.stringify(wantedAttribute)).toBe(expectedJson);
   });
+
+  it('should build a wanted attribute with alternative name', () => {
+    const wantedAttribute = new WantedAttributeBuilder()
+      .withName(TEST_NAME)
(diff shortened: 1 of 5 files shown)


tests/digital_identity_service/policy/wanted.attribute.builder.spec.js
```diff
@@ -131,4 +131,48 @@ describe('WantedAttributeBuilder', () => {
...
+    expect(wantedAttribute).toBeInstanceOf(WantedAttribute);
+    expect(JSON.stringify(wantedAttribute)).toBe(expectedJson);
+  });
+
+  it('should build a wanted attribute with alternative names', () => {
+    const wantedAttribute = new WantedAttributeBuilder()
+      .withName(TEST_NAME)
+      .withAlternativeNames([`${TEST_NAME}-alt1`, `${TEST_NAME}-alt2`])
+      .build();
+
```
abdalmajeed-yoti, line 155: There is still the case of calling `withAlternativeNames` first and then calling `withAlternativeName` after for once or more.

And also the opposite which will only contain the last value if `withAlternativeNames` was called last.

abdalmajeed-yoti (approved):

dhivya-pattabhiraman (approved):

tests/digital_identity_service/policy/wanted.attribute.builder.spec.js
laurent-yoti: Ok, added the first case to the tests.

PR #598, opened 2026-08-19
SDK-2527 Helper method to filter resources per check

laurent-yoti:
Added method "filterForCheck()" in ResourceContainer and method "getResourcesForCheck()"

### DIFF
--- a/src/idv_service/session/retrieve/get.session.result.js
+++ b/src/idv_service/session/retrieve/get.session.result.js
@@ -287,7 +287,10 @@ class GetSessionResult {
   getResourcesForCheck(checkId) {
     const checkResponse = this.checks.find((check) => check.getId() === checkId);
     if (!checkResponse) {
-      throw new Error('Check not found');
+      throw new Error(`Check not found (${checkId})`);
+    }
+    if (!this.resources) {
+      throw new Error('No resources available');
     }
 
     return this.resources.filterForCheck(checkResponse);
--- a/src/idv_service/session/retrieve/get.session.result.js
+++ b/src/idv_service/session/retrieve/get.session.result.js
@@ -279,6 +279,19 @@ class GetSessionResult {
   getDigitalIdShares() {
     return this.digitalIdShares;
   }
+
+  /**
+   * @param checkId {string}
+   * @returns {ResourceContainer}
+   */
+  getResourcesForCheck(checkId) {
+    const checkResponse = this.checks.find((check) => check.getId() === checkId);
+    if (!checkResponse) {
+      throw new Error('Check not found');
+    }
+
+    return this.resources.filterForCheck(checkResponse);
+  }
 }
 
 module.exports = GetSessionResult;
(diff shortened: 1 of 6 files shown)


src/idv_service/session/retrieve/get.session.result.js
```diff
@@ -279,6 +279,19 @@ class GetSessionResult {
...
   }
+
+  /**
+   * @param checkId {string}
+   * @returns {ResourceContainer}
+   */
+  getResourcesForCheck(checkId) {
+    const checkResponse = this.checks.find((check) => check.getId() === checkId);
+    if (!checkResponse) {
+      throw new Error('Check not found');
```
abdalmajeed-yoti: Can we pass `checkId` to the `Error` constructor?

src/idv_service/session/retrieve/resource.container.js
```diff
@@ -148,6 +159,22 @@ class ResourceContainer {
...
+  filterForCheck(checkResponse) {
+    const newResourceContainer = new ResourceContainer({});
+    const resourcesUsed = checkResponse.getResourcesUsed();
+    newResourceContainer.idDocuments = filterResources(this.idDocuments, resourcesUsed);
+    newResourceContainer.supplementaryDocuments = filterResources(this.supplementaryDocuments, resourcesUsed);
+    newResourceContainer.livenessCapture = filterResources(this.livenessCapture, resourcesUsed);
+    newResourceContainer.faceCapture = filterResources(this.faceCapture, resourcesUsed);
+    newResourceContainer.shareCodes = filterResources(this.shareCodes, resourcesUsed);
+    return newResourceContainer;
+  }
```
abdalmajeed-yoti, line 175: Can we refactor it to this?
```js
filterForCheck(checkResponse) {
  const resourcesUsed = checkResponse.getResourcesUsed();
  const filtered = new ResourceContainer({});
  Object.keys(filtered).forEach((key) => {
    filtered[key] = filterResources(this[key], resourcesUsed);
  });
  return filtered;
}

```

src/idv_service/session/retrieve/get.session.result.js
```diff
@@ -279,6 +279,19 @@ class GetSessionResult {
...
+    }
+
+    return this.resources.filterForCheck(checkResponse);
```
abdalmajeed-yoti, line 296: `resources` can be `undefined`.

abdalmajeed-yoti (approved):

khalifah-marfadi (approved):
Other than what @abdalmajeed-yoti mentioned, LGTM

src/idv_service/session/retrieve/get.session.result.js
laurent-yoti: I applied there the same as the Java implementation, I suppose, we can add the `checkId` in - good call.

src/idv_service/session/retrieve/resource.container.js
laurent-yoti: Yeah, I suppose so, it is more bullet-proof - but I am not sure that all keys of a `resourceContainer` instance would be all valid to use.
Again, copied that approach from the Java - I'll have a look

src/idv_service/session/retrieve/get.session.result.js
laurent-yoti: Fair, good catch...!

src/idv_service/session/retrieve/resource.container.js
laurent-yoti: Applied suggested

